PR #31, opened 2015-01-14
Finer grained permissions in resource

leplatrem:
As suggested by @Natim and @ametaireau, just ignore
permissions in AuthorizationPolicy so far.

Once we have FxA scopes for readinglist, we will want
to perform authorization based on token scopes (``readinglist:read``, ``readinglist:write``)...

### DIFF
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -67,7 +67,12 @@ class AuthorizationPolicy(object):
         """Currently we don't check scopes nor permissions.
         Authenticated users only are allowed.
         """
-        return Authenticated in principals
+        PERMISSIONS = {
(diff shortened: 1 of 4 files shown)


readinglist/authentication.py
```diff
@@ -64,12 +64,10 @@ def _get_credentials(self, request):
 @implementer(IAuthorizationPolicy)
 class AuthorizationPolicy(object):
     def permits(self, context, principals, permission):
-        PERMISSIONS = {
-            'articles': Authenticated,
```
Natim: I would rather add `readonly` and `readwrite` here.

leplatrem: Yes if you prefer!

Issue #7, opened 2014-12-22
[API] Avoid having JSON in querystring


Natim:
I've read the API documentation in the README and I would like to avoid this if possible:

    /articles?where={"status": "unread"}

We could change it to:

    /articles?where_status=unread

Or even to:

    /articles?status=unread

If everything is filtering.

The same goes for `?embedded={"devices": 1}`

For the sorting parameter, it would rather use:

    /articles?sort=-date,title

PR #14, opened 2015-01-14
Sort list by attribute

Natim:
Ref #7


setup.py
```diff
@@ -12,6 +12,7 @@
     'pyfxa',
     'pyramid_multiauth',
     'requests',
+    'venusian',
```
Natim: I didn't see where you used it, is it still needed?

leplatrem: I had import errors from pyramid Configurator while running tests under python 3 :(

PR #24, opened 2015-01-16
Error messaging channel

Natim:
* https://github.com/mozilla-services/msisdn-gateway/blob/master/API.md
* https://github.com/mozilla/fxa-oauth-server/blob/master/docs/api.md#errors
* https://docs.services.mozilla.com/loop/apis.html#error-responses

### DIFF
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -1,10 +1,14 @@
+# -*- coding: utf-8 -*-
 from zope.interface import implementer
+
 from pyramid import authentication as base_auth
 from pyramid.interfaces import IAuthenticationPolicy, IAuthorizationPolicy
 from pyramid.security import Authenticated
-from pyramid.httpexceptions import HTTPServiceUnavailable
+
 from fxa.oauth import Client as OAuthClient
 from fxa import errors as fxa_errors
+
+from readinglist.views.errors import service_unavailable
 
 
 def check_credentials(username, password, request):
@@ -58,7 +62,7 @@ class Oauth2AuthenticationPolicy(base_auth.CallbackAuthenticationPolicy):
             profile = auth_client.verify_token(token=auth, scope=scope)
             user_id = profile['user']
         except fxa_errors.OutOfProtocolError:
-            raise HTTPServiceUnavailable()
+            raise service_unavailable()
         except (fxa_errors.InProtocolError, fxa_errors.TrustError):
             return None
         return user_id
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from zope.interface import implementer
 
 from pyramid import authentication as base_auth
@@ -8,7 +7,7 @@ from pyramid.security import Authenticated
 from fxa.oauth import Client as OAuthClient
 from fxa import errors as fxa_errors
 
-from readinglist.views.errors import service_unavailable
+from readinglist.views.errors import HTTPServiceUnavailable
 
 
 def check_credentials(username, password, request):
@@ -62,7 +61,7 @@ class Oauth2AuthenticationPolicy(base_auth.CallbackAuthenticationPolicy):
             profile = auth_client.verify_token(token=auth, scope=scope)
             user_id = profile['user']
         except fxa_errors.OutOfProtocolError:
-            raise service_unavailable()
+            raise HTTPServiceUnavailable()
         except (fxa_errors.InProtocolError, fxa_errors.TrustError):
             return None
         return user_id
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -7,7 +7,7 @@ from pyramid.security import Authenticated
 from fxa.oauth import Client as OAuthClient
 from fxa import errors as fxa_errors
 
-from readinglist.views.errors import HTTPServiceUnavailable
+from readinglist.errors import HTTPServiceUnavailable
 
 
 def check_credentials(username, password, request):
--- a/readinglist/errors.py
+++ b/readinglist/errors.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+import json
+
+from readinglist.utils import Enum
+
+ERRORS = Enum(
+    MISSING_AUTH_TOKEN=104,
+    INVALID_AUTH_TOKEN=105,
+    BADJSON=106,
+    INVALID_PARAMETERS=107,
+    MISSING_PARAMETERS=108,
+    INVALID_POSTED_DATA=109,
+    INVALID_RESOURCE_ID=110,
+    MISSING_RESOURCE=111,
+    FORBIDDEN=121,
+    REQUEST_TOO_LARGE=113,
+    CLIENT_REACHED_CAPACITY=117,
+    UNDEFINED=999,
+    BACKEND=201
+)
+
+
+def get_formatted_error(code, errno, error, message=None, info=None):
+    result = {
+        "code": code,
+        "errno": errno,
+        "error": error
+    }
+
+    if message is not None:
+        result['message'] = message
+
+    if info is not None:
+        result['info'] = info
+
+    return json.dumps(result)
--- a/readinglist/errors.py
+++ b/readinglist/errors.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 import json
 
 from readinglist.utils import Enum
--- a/readinglist/errors.py
+++ b/readinglist/errors.py
@@ -1,6 +1,10 @@
 import json
 
 from readinglist.utils import Enum
+from pyramid.config import global_registries
+from pyramid.httpexceptions import (
+    HTTPServiceUnavailable as PyramidHTTPServiceUnavailable
+)
 
 ERRORS = Enum(
     MISSING_AUTH_TOKEN=104,
@@ -33,3 +37,26 @@ def get_formatted_error(code, errno, error, message=None, info=None):
         result['info'] = info
 
     return json.dumps(result)
+
+
+class HTTPServiceUnavailable(PyramidHTTPServiceUnavailable):
+    """Return an HTTPServiceUnavailable formatted error."""
+
+    def __init__(self, **kwargs):
+        if 'body' not in kwargs:
+            kwargs['body'] = get_formatted_error(
+                503, ERRORS.BACKEND, "Service unavailable",
+                "Service unavailable due to high load, please retry later.")
+
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/json'
+
+        if 'headers' not in kwargs:
+            kwargs['headers'] = []
+
+        settings = global_registries.last.settings
+        kwargs['headers'].append(
+            ("Retry-After", settings.get('readinglist.retry_after', "30"))
+        )
+
+        super(HTTPServiceUnavailable, self).__init__(**kwargs)
--- a/readinglist/errors.py
+++ b/readinglist/errors.py
@@ -1,10 +1,8 @@
-import json
 from pyramid.config import global_registries
 from pyramid.httpexceptions import (
-    HTTPServiceUnavailable as PyramidHTTPServiceUnavailable, HTTPError
+    HTTPServiceUnavailable as PyramidHTTPServiceUnavailable, HTTPBadRequest
 )
-from pyramid.response import Response
-from readinglist.utils import Enum
+from readinglist.utils import Enum, json
 
 ERRORS = Enum(
     MISSING_AUTH_TOKEN=104,
@@ -55,25 +53,13 @@ class HTTPServiceUnavailable(PyramidHTTPServiceUnavailable):
             kwargs['headers'] = []
 
         settings = global_registries.last.settings
+        retry_after = settings.get(
+            'readinglist.retry_after', "30").encode("utf-8")
         kwargs['headers'].append(
-            ("Retry-After", settings.get('readinglist.retry_after', "30"))
+            ("Retry-After", retry_after)
         )
 
         super(HTTPServiceUnavailable, self).__init__(**kwargs)
-
-
-class _JSONError(HTTPError):
-    def __init__(self, errors, status=400):
-        try:
-            code = int(status)
-        except ValueError:
-            code = int(status[:3])
-
-        body = get_formatted_error(
-            code=code, errno=ERRORS.INVALID_PARAMETERS,
-            error="Invalid parameters",
-            message='%(name)s in %(location)s: %(description)s' % errors[0])
-        super(Response, self).__init__(body, content_type='application/json', status=status)
 
 
 def json_error(errors):
@@ -81,4 +67,19 @@ def json_error(errors):
 
     The HTTP error content type is "application/json"
     """
-    return _JSONError(errors, errors.status)
+    sorted_errors = sorted(errors, key=lambda x: "%s" % x['name'])
+    error = sorted_errors[0]
+    if error['name'] is not None:
+        if error['name'] in error['description']:
+            message = error['description']
+        else:
+            message = '%(name)s in %(location)s: %(description)s' % error
+    else:
+        message = '%(location)s: %(description)s' % error
+    body = get_formatted_error(
+        code=400, errno=ERRORS.INVALID_PARAMETERS,
+        error="Invalid parameters",
+        message=message)
+    response = HTTPBadRequest(body=body, content_type='application/json')
+    response.status = errors.status
+    return response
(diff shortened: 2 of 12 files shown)


Natim:
@leplatrem r?

readinglist/authentication.py
```diff
@@ -1,11 +1,15 @@
+# -*- coding: utf-8 -*-
```
leplatrem: Why would we need that ?

readinglist/authentication.py
```diff
@@ -55,7 +59,7 @@ def _get_credentials(self, request):
             profile = auth_client.verify_token(token=auth, scope=scope)
             user_id = profile['user']
         except fxa_errors.OutOfProtocolError:
-            raise HTTPServiceUnavailable()
+            raise service_unavailable()
```
leplatrem: I find it strange to use a function here. Why not subclassing ?

readinglist/tests/test_views_errors.py
```diff
@@ -0,0 +1,53 @@
...
+import mock
+
+from fxa import errors as fxa_errors
+
+from readinglist.errors import ERRORS
+from .support import BaseWebTest
+
+
+class ErrorViewTest(BaseWebTest, unittest.TestCase):
+
```
leplatrem: What about 400 validations errors ?

readinglist/views/errors.py
```diff
@@ -9,9 +18,57 @@ def authorization_required(request):
...
+            ERRORS.MISSING_RESOURCE,
+            "Not Found",
+            "The resource your are looking for could not be found."),
+        content_type='application/json')
     return response
+
+
+@view_config(context=Exception)
+def error(context, request):
+    """Display an error message and record it in Sentry."""
```
leplatrem: Sentry ?

readinglist/views/errors.py
```diff
@@ -9,9 +18,57 @@ def authorization_required(request):
...
+    # client = Client()
+    try:
+        raise
+    except Exception:
+        # client.captureException()
+        traceback.print_exc()
```
leplatrem: print ?

readinglist/views/errors.py
```diff
@@ -9,9 +18,57 @@ def authorization_required(request):
...
+    return HTTPInternalServerError(
+        body=get_formatted_error(
+            500,
+            ERRORS.UNDEFINED,
+            "Internal Server Error",
+            "A programmatic error occured, developers have been informed."),
+        content_type='application/json')
+
+
+def service_unavailable():
```
leplatrem: See above comment about subclassing :)

leplatrem: Maybe that's the way to do it in Pyramid :)

readinglist/views/heartbeat.py
```diff
@@ -16,7 +16,6 @@ def get_heartbeat(request):
     status = dict(database=database)
     has_error = not all([v for v in status.values()])
     if has_error:
-        # XXX: use Cornice errors
         request.response.status = 503
```
leplatrem: Is this enough for cornice to return our json error responses ?

leplatrem:
I would have thought that relying on exceptions would have been easier... 

Like raise an exception [1] and catch it in a common decorator to return a formatted error. But I am absolutely not confident about what would be the best approach :) It's just for the pleasure of arguing :)

[1] https://github.com/mozilla/PyFxA/blob/master/fxa/errors.py#L26-L41 for instance

readinglist/views/heartbeat.py
Natim: We don't want heartbeat endpoint to return a JSON error but just the heartbeat error format.

readinglist/views/errors.py
Natim: This is another ticket, I let the comments that helps to implement it.

readinglist/views/errors.py
Natim: To be replace with sentry.

readinglist/tests/test_views_errors.py
Natim: 400 validation errors are handled differently. We still have to work on this.

readinglist/authentication.py
Natim: Yes we should probably subclass.

readinglist/authentication.py
Natim: To tell Python that the file is in utf-8, it is mostly a best practice nothing that we really need for this specific file.

Natim:
I like it to be explicit.

readinglist/views/errors.py
Natim: I will try to bind it to stderr.

readinglist/errors.py
```diff
@@ -0,0 +1,59 @@
...
+                503, ERRORS.BACKEND, "Service unavailable",
+                "Service unavailable due to high load, please retry later.")
+
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/json'
+
+        if 'headers' not in kwargs:
+            kwargs['headers'] = []
+
+        kwargs['headers'].append(("Retry-After", 30))
```
Natim: I'd like to load the 30 value from the configuration. Any idea of how I can manage to do so?

readinglist/errors.py
```diff
@@ -0,0 +1,62 @@
...
+        "code": code,
+        "errno": errno,
+        "error": error
+    }
+
+    if message is not None:
+        result['message'] = message
+
+    if info is not None:
+        result['info'] = info
```
leplatrem: Info was not used, should we add a link to the API design proposal in errors responses ?

Natim: Actually the protocol say that this field can contain an URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1120671#c8

readinglist/resource.py
```diff
@@ -151,3 +152,10 @@ def delete(self):
         record_id = self.request.matchdict['id']
         self.record = self.db.delete(record_id=record_id, **self.db_kwargs)
         return self.record
+
+
+def resource(*args, **kwargs):
+    if 'error_handler' not in kwargs:
+        kwargs['error_handler'] = json_error
+
+    return cornice_resource(*args, **kwargs)
```
leplatrem: yes, well this will be taken care of in PR #34

readinglist/errors.py
```diff
@@ -0,0 +1,84 @@
...
+        try:
+            code = int(status)
+        except ValueError:
+            code = int(status[:3])
+
+        body = get_formatted_error(
+            code=code, errno=ERRORS.INVALID_PARAMETERS,
+            error="Invalid parameters",
+            message='%(name)s in %(location)s: %(description)s' % errors[0])
+        super(Response, self).__init__(body, content_type='application/json', status=status)
```
leplatrem: bypass HTTPError class in super() ?

readinglist/errors.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, errors, status=400):
+        try:
+            code = int(status)
+        except ValueError:
+            code = int(status[:3])
+
+        body = get_formatted_error(
+            code=code, errno=ERRORS.INVALID_PARAMETERS,
+            error="Invalid parameters",
+            message='%(name)s in %(location)s: %(description)s' % errors[0])
```
leplatrem: I believe we can add a ``validation`` attribute in error responses (see API proposal)

readinglist/errors.py
```diff
@@ -0,0 +1,84 @@
...
+
+class _JSONError(HTTPError):
+    def __init__(self, errors, status=400):
+        try:
+            code = int(status)
+        except ValueError:
+            code = int(status[:3])
+
+        body = get_formatted_error(
+            code=code, errno=ERRORS.INVALID_PARAMETERS,
```
leplatrem: split lines ?

readinglist/errors.py
```diff
@@ -0,0 +1,84 @@
...
+
+        super(HTTPServiceUnavailable, self).__init__(**kwargs)
+
+
+class _JSONError(HTTPError):
+    def __init__(self, errors, status=400):
+        try:
+            code = int(status)
+        except ValueError:
+            code = int(status[:3])
```
tarekziade: if JSONError takes an int for a status, this fallback is a bit of a hack. probably unecessary

Natim: I had to do it after writting the test.

tarekziade: according to the pyramid code, status is created by the Response class by concatenating the int "code" and the str "line" - so I guess you could have "code=400" here instead, drop the custom code, and let Response build status.

If a caller uses status instead of code on _JsonError, it's a bug

readinglist/errors.py
```diff
@@ -0,0 +1,84 @@
...
+
+        if 'content_type' not in kwargs:
+            kwargs['content_type'] = 'application/json'
+
+        if 'headers' not in kwargs:
+            kwargs['headers'] = []
+
+        settings = global_registries.last.settings
+        kwargs['headers'].append(
+            ("Retry-After", settings.get('readinglist.retry_after', "30"))
```
leplatrem: Should come from settings ?

readinglist/resource.py
Natim: Do you know how this should be rebased on top of the new master?

readinglist/errors.py
Natim: I know, the code is coming from Cornice: https://github.com/mozilla-services/cornice/blob/master/cornice/util.py#L88

But we can probably extend resource directly?

readinglist/errors.py
Natim: We talked about it with Richard and Chris and they both agreed that we should return only one error at a time and try to follow the same error protocol for all services. That's the reason why I choose to do it like so.

readinglist/errors.py
Natim: Yes it does: `settings.get('readinglist.retry_after'`

readinglist/resource.py
leplatrem: Yes, in the crud decorator:
```python
        params = dict(collection_path='/{0}s'.format(resource_name),
                      path='/{0}s/{{id}}'.format(resource_name),
                      description='Collection of {0}'.format(resource_name),
                      error_handler=json_error,
                      depth=2)
```

readinglist/views/errors.py
```diff
@@ -9,9 +18,49 @@ def authorization_required(request):
...
+
+
+@notfound_view_config()
+def page_not_found(request):
+    """Return a JSON 404 error page."""
+    response = HTTPNotFound(
+        body=get_formatted_error(
+            404,
+            ERRORS.MISSING_RESOURCE,
+            "Not Found",
```
leplatrem: Identation not consistent with 401 and 403 :)

readinglist/errors.py
leplatrem: Yes ok, but it is not used at all yet (see coverage)

readinglist/errors.py
leplatrem: Shouldn't you encode it btw ?
https://github.com/mozilla-services/readinglist/pull/17/files#diff-d4d5d6d29138ae34862fef2610ec0429R60

readinglist/errors.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        super(HTTPServiceUnavailable, self).__init__(**kwargs)
+
+
+class _JSONError(HTTPError):
+    def __init__(self, errors, status="400 BAD REQUEST"):
```
leplatrem: Looks like Pyramid have capitalized status by default
https://github.com/Pylons/pyramid/blob/master/pyramid/httpexceptions.py#L565-L566

readinglist/errors.py
```diff
@@ -0,0 +1,80 @@
...
+        body = get_formatted_error(
+            code=400, errno=ERRORS.INVALID_PARAMETERS,
```
tarekziade: If I change the status to "403 forbidden", you are forcing the body to 400 here. It would be simpler to let the  base class take care of computing the code given the status, and override the body *after* that.

Natim: Yes it is what I did before and you asked me to change it there. Also we should always use this error with a 400 because of the errno.

tarekziade: wat ? I asked you to be consistent - your initial change was setting the status to an int and baking the code by extracting the status if it was a string, then you moved to this version where we can change the status but the code is forced to 400. 

What's the goal here ? can I change the status and/or the code or do you force  it to 400 ?  what "because of the errno" means ?

tarekziade: If your goal is to always return a 400 with a json body, you can probably drop your custom class and do something like this:

```python

def json_error(errors):
    body = get_formatted_error(code=400, errno=ERRORS.INVALID_PARAMETERS,
                                                 error="Invalid parameters", 
                                                 message='%(name)s in %(location)s: %(description)s' % errors[0])
    response = HTTPBadRequest(body=body)
    response.content_type = 'application/json'
    return response
```

unless you want people to be able to set the code like you initially did, but it seems not?

Natim: Yes you're probably right.

Natim:
PR updated.

readinglist/errors.py
Natim: I added a test to cover this.

tox.ini
```diff
@@ -6,6 +6,7 @@ commands =
     python --version
     nosetests --with-coverage --cover-package=readinglist readinglist {posargs}
 deps =
+	simplejson
```
leplatrem: Indentation ?

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
+from .utils import json
```
ametaireau: we should be consistent between relaltive and absolute imports. Here you're doing both.

Natim: Yes my bad.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+from pyramid.config import global_registries
+from pyramid.httpexceptions import (
+    HTTPServiceUnavailable as PyramidHTTPServiceUnavailable, HTTPBadRequest
+)
+from readinglist.utils import Enum
```
ametaireau: order the imports alphabetically.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+    MISSING_RESOURCE=111,
+    FORBIDDEN=121,
+    REQUEST_TOO_LARGE=113,
+    CLIENT_REACHED_CAPACITY=117,
+    UNDEFINED=999,
+    BACKEND=201
+)
+
+
+def get_formatted_error(code, errno, error, message=None, info=None):
```
ametaireau: Please add a comment.

ametaireau: maybe we should name this `format_error` instead.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+            'readinglist.retry_after', "30").encode("utf-8")
+        kwargs['headers'].append(
+            ("Retry-After", retry_after)
+        )
+
+        super(HTTPServiceUnavailable, self).__init__(**kwargs)
+
+
+def json_error(errors):
+    """Returns an HTTPError with the given status and message.
```
ametaireau: "Return", use the infinitive for consistency

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+
+    The HTTP error content type is "application/json"
+    """
+    sorted_errors = sorted(errors, key=lambda x: "%s" % x['name'])
```
ametaireau: I don't get why we need to do the "%s" trick here. wouldn't `sorted(errors, key=lambda x: x['name'])` be enough? or if we really need to convert to a string, then use `str`.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+    error = sorted_errors[0]
```
ametaireau: Before doing that we should check the length of the list.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+    if error['name'] is not None:
```
ametaireau: set name = error['name'] beforehand to ease reading.

readinglist/errors.py
```diff
@@ -0,0 +1,86 @@
...
+        if error['name'] in error['description']:
+            message = error['description']
+        else:
+            message = '%(name)s in %(location)s: %(description)s' % error
+    else:
+        message = '%(location)s: %(description)s' % error
```
ametaireau: Please add an empty line here.

readinglist/errors.py
Natim: Sometimes x['name'] is None and Python3 cannot compare None and str.

readinglist/errors.py
Natim: Yes we could also we will never enter this json_error without any error.

PR #17, opened 2015-01-19
Backoff indicators

Natim:
* http://docs.services.mozilla.com/token/apis.html?highlight=backoff#request-headers


Natim:
@leplatrem r?

leplatrem:
LGTM to me :)

@Natim, please add an empty value (or commented line) in settings so that we can keep track of this backoff value

ametaireau:
LGTM as well.

PR #47, opened 2015-01-19
Rework load tests

leplatrem:
* [x] Update to work with Cornice version
* [x] Add randomness
* [x] Run on Travis
* [x] Improve assertions ?


Makefile
```diff
@@ -36,4 +37,10 @@ tests:
 	tox
 
 bench: install-dev
-	$(VENV)/bin/loads-runner --config=./loadtests/bench.ini --server-url=$(SERVER_URL) loadtests.TestPOC.test_all
+	$(VENV)/bin/loads-runner --config=./loadtests/bench.ini --server-url=$(SERVER_URL) loadtests.TestBasic.test_all
+
+loadtests-check:
+	make serve & PID=$$!; \
+	  sleep 1; \
+	  make bench; \
```
Natim: It should be make test if it is a smoketest.

Makefile
```diff
@@ -36,4 +37,10 @@ tests:
 	tox
 
 bench: install-dev
-	$(VENV)/bin/loads-runner --config=./loadtests/bench.ini --server-url=$(SERVER_URL) loadtests.TestPOC.test_all
+	$(VENV)/bin/loads-runner --config=./loadtests/bench.ini --server-url=$(SERVER_URL) loadtests.TestBasic.test_all
+
+loadtests-check:
+	make serve & PID=$$!; \
```
tarekziade: maybe we could put make server on its own line, so in case it fails to start we don't run the load tests.

Makefile
```diff
@@ -1,4 +1,5 @@
 SERVER_URL = http://localhost:8000
```
ametaireau: Just a remark: for all the other projects, we kept the loadtest makefile target separated from the main makefile (e.g. a different Makefile file inside the loadtests directory).

I think I like it best to have everything in here, but we should try to be consistent in the future.

loadtests/__init__.py
```diff
@@ -1,54 +1,169 @@
...
+        super(TestBasic, self).__init__(*args, **kwargs)
+
+        self.random_user = None
 
-        self.alice_auth = HTTPBasicAuth('alice', 'secret')
-        self.john_auth = HTTPBasicAuth('john', 'secret')
+        # Create at least one record per user
+        for user in self.available_users:
+            self.basic_auth = HTTPBasicAuth(*user)
+            self.test_create_record()
```
ametaireau: I'm not sure there is a value to prefixing the actions by `test`.

loadtests/__init__.py
```diff
@@ -1,54 +1,169 @@
...
     def api_url(self, path):
-        return "{0}/v1/{1}".format(self.server_url, path)
+        return "{0}/v0/{1}".format(self.server_url, path)
 
     def test_all(self):
-        self.test_home()
-        self.test_create_record()
-        self.test_update_status()
-        self.test_list_filtered()
+        if random.randint(0, 100) < PERCENTAGE_OF_CREATION:
```
ametaireau: I would create an intermediate method to check if we should do this test or not. I probably also would rework all these constants in a python dict and have the key be the subject, so we could do something like `if self.should_run('creation'):`

Makefile
ametaireau: Also, we have theorically different targets which are defined by [configuration files](https://github.com/mozilla-services/loop-server/tree/master/loadtests/config), we should continue to do the same thing I believe :)

Makefile
Natim: We cannot because we want to stop the server if the test passed or didn't.

loadtests/__init__.py
```diff
@@ -1,54 +1,142 @@
...
     def test_all(self):
-        self.test_home()
-        self.test_create_record()
-        self.test_update_status()
-        self.test_list_filtered()
-
-    def test_home(self):
-        res = self.session.get(self.api_url('articles'), auth=self.alice_auth)
-        self.assertEqual(res.status_code, 200)
+        for action, percentage in ACTIONS_PERCENTAGE:
```
tarekziade: I am confused here because a single call is going to generate one to nine tests. Is this the intended behavior ?

leplatrem: Ok, got it. That's fixed :)

tarekziade:
lgtm

loadtests/__init__.py
```diff
@@ -1,54 +1,145 @@
...
+        super(TestBasic, self).__init__(*args, **kwargs)
+
+        self.random_user = uuid.uuid4().hex
+        self.basic_auth = HTTPBasicAuth(self.random_user, 'secret')
 
-        self.alice_auth = HTTPBasicAuth('alice', 'secret')
-        self.john_auth = HTTPBasicAuth('john', 'secret')
+        # Create at least some records for this user
+        total_records = random.randint(3, 100)
+        for i in range(total_records):
```
Natim: six.range ?

Natim: Maybe the answer is no here.

Natim:
Except for the separate loadtest Makefile it looks good to me too.

Natim:
Loadtest are not passing on travis, see https://travis-ci.org/mozilla-services/readinglist/jobs/47801450

leplatrem:
The current error could be related with having both virtualenvs activated, no ?

Natim:
Not it is just because loads doesn't install correctly.

PR #48, opened 2015-01-20
Return 201 when creation is successful

leplatrem:


readinglist/resource.py
```diff
@@ -176,6 +176,7 @@ def collection_post(self):
         new_record = self.request.validated
         new_record = self.process_record(new_record)
         self.record = self.db.create(record=new_record, **self.db_kwargs)
+        self.request.response.status = '201 Created'
```
tarekziade: I am not sure why we want to mess with status messages in our code base. This is to be dealt by the framework. We shoud set a 201 code and the thing should set the status automatically

tarekziade: see https://github.com/Pylons/webob/blob/master/webob/response.py#L274

```python
self.request.response.status_code = 201
```

readinglist/resource.py
```diff
@@ -33,7 +33,7 @@ def wrapped_view(self, *args, **kwargs):
                 # Transform the errors we got from colander into cornice errors
                 for field, error in e.asdict().items():
                     self.request.errors.add('body', field, error)
-            self.request.errors.status = "400 Bad Request"
+            self.request.errors.status = 400
```
ametaireau: why not `status_code` here?

readinglist/resource.py
```diff
@@ -18,7 +18,7 @@ def wrapped_view(self, *args, **kwargs):
                 return view(self, *args, **kwargs)
             except RecordNotFoundError as e:
                 self.request.errors.add('path', 'id', six.text_type(e))
-                self.request.errors.status = "404 Resource Not Found"
+                self.request.errors.status = 404
```
ametaireau: why not `status_code` here?

Natim: Because it is request.errors and not response.status_code

Natim: This is the normal Cornice behavior: http://cornice.readthedocs.org/en/latest/validation.html#changing-the-status-code-from-validators

Natim:
LGTM

PR #49, opened 2015-01-20
Return 400 on invalid JSON

leplatrem:


readinglist/resource.py
```diff
@@ -29,6 +29,8 @@ def wrap(view):
         def wrapped_view(self, *args, **kwargs):
             try:
                 return view(self, *args, **kwargs)
+            except ValueError as e:
+                self.request.errors.add('body', 'body', six.text_type(e))
```
ametaireau: `name="body"` ? I feel this is kinda weird.

Natim: It should be `name=None`

readinglist/tests/support.py
```diff
@@ -109,6 +109,13 @@ def test_invalid_record_raises_error(self):
                                   status=400)
         self.assertIn('errors', resp.json)
 
+    def test_create_with_invalid_json(self):
+        body = "{'foo': 'bad json'}"
```
ametaireau: We should use something more obvious than the invalid `'`, because nobody is aware of that. Adding a missing closing bracket or something like that would be more effective.

readinglist/tests/support.py
```diff
@@ -109,6 +109,13 @@ def test_invalid_record_raises_error(self):
...
+        self.app.post(self.collection_url,
+                      body,
+                      headers=self.headers,
+                      status=400)
```
ametaireau: We probably should test what's the content of the 400

Natim: It will be the case with #24 also there is an helper to test  that `assertFormattedError`

Natim: Now that #24 has been merged, I think we can add the use of assertFormattedError here.

PR #50, opened 2015-01-20
Redis backend

ametaireau:
This is the very begining of the redis backend class.

It's not finished, but since I'll be on PTO until EOW, I figured out it would be easier to have it around to continue the work.
I haven't used limpyd because it seems to not be very relevant for us, after having a look at how it works.


Natim:
Redis is missing in the requirements.

ametaireau:
yeah. Functional tests (for the views) aren't passing anyway

readinglist/tests/test_backend.py
```diff
@@ -29,3 +34,93 @@ def test_mandatory_overrides(self):
         ]
         for call in calls:
             self.assertRaises(NotImplementedError, *call)
+
+
+class TestResource(object):
+    id_field = "id"
```
leplatrem: This is what I want to do in a ``test_resource.py``

leplatrem:
I'm wondering if we should run all tests on Redis as well (like we do for Daybed), or at least be able to do it in a specific task on TravisCI

Natim:
I think we should test all backends.

conf/readinglist.ini
```diff
@@ -1,7 +1,7 @@
 [app:main]
 use = egg:readinglist
 pyramid.debug_notfound = true
-readinglist.backend = readinglist.backend.memory
+readinglist.backend = readinglist.backend.simpleredis
```
leplatrem: This means that the README deserves a bit more of documentation :)

I suggest we leave memory by default but give details about changing conf when running Redis

README.rst
```diff
@@ -17,11 +17,58 @@ API
...
+
+Install Redis
+=============
+
+Linux
+-----
+
+On debian / ubuntu based systems:::
+
+    sh
```
Natim: What is this `sh` thing ? I would rather put a $ in front of the following line or use 

    .. code-block:: sh

README.rst
```diff
@@ -17,11 +17,58 @@ API
...
+    apt-get install redis-server
+
+
+or:::
+
+    sh
```
Natim: ditto

PR #52, opened 2015-01-21
Checking filter and sort querystring (fixes #44)

### LINKED ISSUE #44: Raise errors when sorting/filtering fields are unknown
Related #13, #43

leplatrem:


leplatrem:
Rebased, @Natim r?

readinglist/resource.py
```diff
@@ -131,6 +131,24 @@ def _extract_filters(self, queryparams):
...
+                    }
+                    self.request.errors.add(**error_details)
+                    return
+
+                filters.append(
+                    (self.modified_field, value, COMPARISON.MIN)
+                )
+                continue
+
             m = re.match(r'^(min|max|not)_(\w+)$', param.strip())
```
Natim: I think we should do the `param.strip()` at the top then.

readinglist/resource.py
```diff
@@ -129,26 +129,44 @@ def _extract_filters(self, queryparams):
...
+                    error_details = {
+                        'name': param,
+                        'location': 'querystring',
+                        'description': 'Invalid value for _since'
+                    }
+                    self.request.errors.add(**error_details)
+                    return
+
+                filters.append(
+                    (self.modified_field, value, COMPARISON.MIN)
```
ametaireau: Should we use [names tuples](https://docs.python.org/2/library/collections.html#collections.namedtuple) instead of triples?

PR #53, opened 2015-01-21
Prevent to modify readonly fields (fixes #26)

### LINKED ISSUE #26: Make sure some attributes cannot be changed by the client.
Some attributes are immutable:

* ``url``
* ``stored_on``
* ~~``resolved_url``~~
* ~~``resolved_title``~~
* ``last_modified``

leplatrem:


readinglist/resource.py
```diff
@@ -93,6 +95,12 @@ class RessourceSchema(colander.MappingSchema):
     _id = colander.SchemaNode(colander.String(), missing=colander.drop)
     last_modified = TimeStamp()
 
+    __readonly_fields__ = ('_id', 'last_modified')
```
tarekziade: why do you use double-dunder ? is it something from Colander ? if not it's potential trouble when accessing it

Natim: Yes I had the same strange feeling reading this.

leplatrem: Just a naive inspiration to distinguish this from actual schema fields :/

Any alternative idea ?


```python
class Schema(...):
    field1 = ...
    field2 = ...

    class Options:
        readonly_fields = ...
```

Natim:
```
class Schema(...):
    field1 = ...
    field2 = ...

    class Meta:
        readonly_fields = ...
```

Looks like Django ideas are coming.

Maybe just `_readonly_fields`?

readinglist/errors.py
```diff
@@ -92,3 +92,11 @@ def json_error(errors):
     response.status = errors.status
 
     return response
+
+
+class ImmutableFieldError(Exception):
+    """Raised when attempting to modify a read-only field."""
+    def __init__(self, field, msg=None):
+            self.field = field
```
ametaireau: It seems we have too many spaces here.

readinglist/resource.py
ametaireau: I actually quite like the `__readonly_fields__`, I find its meaning obvious, and says to me "this is a metadata field". I don't know where this is coming from though :D

ametaireau: Having an `Options` or `Meta` subclass seems also good to me, and it seems easy to have interchangeable subclasses (with different readonly_fields for instance, or which computes the fields only when accessed).

PR #54, opened 2015-01-21
Support of min, max, not keywords in filters (fixes #43)

### LINKED ISSUE #43: Add min, max, not syntax for filters

leplatrem:
Will have to be rebased on top of #52

### DIFF
--- a/readinglist/resource.py
+++ b/readinglist/resource.py
@@ -130,9 +130,19 @@ class BaseResource(object):
 
         for param, value in queryparams.items():
             value = native_value(value)
-            if param in self.known_fields:
-                filters.append((param, value, COMPARISON.EQ))
-            if param == '_since':
+
+            m = re.match(r'\s?(min|max|not)_(\w+)\s?', param)
+
+            if m:
+                keyword, field = m.groups()
+                operator = COMPARISON[keyword]
+            else:
+                operator, field = COMPARISON.EQ, param
+
+            if field in self.known_fields:
+                filters.append((field, value, operator))
+
+            if field == '_since':
                 if isinstance(value, six.integer_types):
                     filters.append(
                         (self.modified_field, value, COMPARISON.MIN)
--- a/readinglist/resource.py
+++ b/readinglist/resource.py
@@ -131,13 +131,12 @@ class BaseResource(object):
         for param, value in queryparams.items():
             value = native_value(value)
 
-            m = re.match(r'\s?(min|max|not)_(\w+)\s?', param)
-
+            m = re.match(r'^(min|max|not)_(\w+)$', param.strip())
             if m:
                 keyword, field = m.groups()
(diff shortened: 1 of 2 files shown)


readinglist/tests/test_views_article.py
```diff
@@ -82,6 +82,21 @@ def test_string_filters_naively_by_value(self):
...
+    def test_minimal_value(self):
+        resp = self.app.get('/articles?min_status=2', headers=self.headers)
+        values = [item['status'] for item in resp.json['items']]
+        self.assertTrue(all([value >= 2 for value in values]))
+
+    def test_maximal_value(self):
+        resp = self.app.get('/articles?max_status=2', headers=self.headers)
+        values = [item['status'] for item in resp.json['items']]
+        self.assertTrue(all([value <= 2 for value in values]))
+
```
Natim: I would add a test to make sure `madmax_status=2` works properly.

readinglist/resource.py
```diff
@@ -127,9 +127,23 @@ def _extract_filters(self, queryparams):
 
         for param, value in queryparams.items():
             value = native_value(value)
-            if param in self.known_fields:
-                filters.append((param, value, '=='))
-            if param == '_since':
+
+            m = re.match(r'\s?(min|max|not)_(\w+)\s?', param)
+            if m:
+                keyword, field = m.groups()
```
tarekziade: smells like a bunch of constants to me.  min, max, not, <= etc..

readinglist/tests/test_views_article.py
leplatrem: ``madmax``?

Natim:
```
    def test_maximal_false_positive(self):
        resp = self.app.get('/articles?madmax_status=2', headers=self.headers)
        values = [item['status'] for item in resp.json['items']]
        self.assertFalse(all([value <= 2 for value in values]))
```

Natim: This regexp `re.match(r'\s?(min|max|not)_(\w+)\s?', param)` is too permisive.

leplatrem: ``/articles?madmax_status=2`` will fail with PR #44 since this field will be unknown.

If the regexp is too permissive, prove it with a test :)

leplatrem: It lacks start and end markers indeed :) ``r'^\s?(min|max|not)_(\w+)\s?$'``

Natim: Yes I did it like so: https://github.com/mozilla-services/readinglist/pull/54/files#diff-54d0a7a5d054e03f91e5dec6de832ae4R131

Natim:
Need to be rebase on top of #58 in order to benefit from the constant defined there.

Natim:
Updated. r? @leplatrem

PR #55, opened 2015-01-21
Increase article titles length (fixes #37)

### LINKED ISSUE #37: Increase length of title and resolved_title to 1024

leplatrem:


readinglist/views/article.py
```diff
@@ -14,14 +14,29 @@ def preparer(self, appstruct):
         return strip_whitespace(appstruct)
 
 
+class URL(SchemaNode):
+    """String representing a URL."""
+    schema_type = String
+    validator = colander.All(colander.url, colander.Length(min=1, max=2048))
```
Natim: The max should be a configuration.

readinglist/views/article.py
```diff
@@ -14,14 +14,29 @@ def preparer(self, appstruct):
...
+
+    def preparer(self, appstruct):
+        return strip_whitespace(appstruct)
+
+
+class ArticleTitle(SchemaNode):
+    """String representing the title of an article."""
+    schema_type = String
+    validator = colander.Length(min=1, max=1024)
```
Natim: The max should be a configuration.

readinglist/views/article.py
leplatrem: Changing the value might imply running some kind of migration depending on the backend. Should we allow that from settings?

Using a constant in code base seems enough IHMO

Putting configuration in schemas is painful :)

Natim: > putting configuration in schemas is painful

Yes I have seen that.

ametaireau: open for discussion: should we truncate the URL when it's longer than what we want? (rather than rejecting it)

leplatrem: this question could be handled in a separate issue, since it also concerns article titles (@rnewman proposition)

Natim: Agreed

PR #57, opened 2015-01-22
Switch to vector-clock timestamps (fixes #4)

leplatrem:
Avoid races using vector-clock timestamps, as described in #4 : 

```
Client A: [        fetch                 ] t = 1234
Client B:                     [insert] t = 1230
Client A:                                           [fetch since t=1234] t=1345
```

Basically, a naive approach: 
* build millisecond timestamps with incremented fraction
* each incoming request has a different timestamp (even if less than 1msec of difference)
* ``last_modified`` is assigned with this request timestamp
* request timestamp is provided in request response

To be improved/reviewed/discussed:
* Filter ``_since`` is exclusive (strictly greater ``>``)
* ``stored_on`` is not receiving the same value as ``last_modified``
* Implementation of timestamping shall be implemented by storage backend (guarantee timestamp unicity in *multiprocess* environment)

### DIFF
--- a/readinglist/utils.py
+++ b/readinglist/utils.py
@@ -4,10 +4,17 @@ except ImportError:  # pragma: no cover
     import json  # NOQA
 
 import ast
+import decimal
+import threading
+import time
+
 from colander import null
+
 
 # removes whitespace, newlines, and tabs from the beginning/end of a string
 strip_whitespace = lambda v: v.strip(' \t\n\r') if v is not null else v
+
+msec_time = lambda: int(time.time() * 1000.0)  # floor
 
 
 def native_value(value):
@@ -34,3 +41,39 @@ COMPARISON = Enum(
     EQ='==',
     GT='>',
 )
+
+
+class TimeStamper(object):
+    """A helper to generate accurate timestamps.
+
+    It increments a fraction in case prevent two simultaneous calls to return
+    the same timestamp.
+
+    This is mainly used for in-memory backend. Ideally, this task should rely
+    on storage backend transaction capabilities to guarantee timestamp unicity
+    in a multiprocess environment.
+
+    .. note:
+
+        Original comes from http://stackoverflow.com/a/157711
+    """
+    fraction = decimal.Decimal('0.001')
+
+    def __init__(self):
+        self.lock = threading.Lock()
+        self.prev = None
+        self.count = 0
+
+    def now(self):
+        with self.lock:
+            timestamp = decimal.Decimal(msec_time())
+            if timestamp == self.prev:
+                timestamp += self.count * self.fraction
+                self.count += 1
+            else:
+                self.prev = timestamp
+                self.count = 1
+            return timestamp
+
+
+timestamper = TimeStamper()
--- a/readinglist/utils.py
+++ b/readinglist/utils.py
@@ -4,8 +4,6 @@ except ImportError:  # pragma: no cover
     import json  # NOQA
 
 import ast
-import decimal
-import threading
 import time
 
 from colander import null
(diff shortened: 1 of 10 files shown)


tarekziade:
> each incoming request has a different timestamp (even if less than 1msec of difference)

so what happens if I have 50 reqs at the same ms ? do you move timestamps in the future ?

readinglist/utils.py
```diff
@@ -24,3 +31,39 @@ def native_value(value):
...
+    in a multiprocess environment.
+
+    .. note:
+
+        Original comes from http://stackoverflow.com/a/157711
+    """
+    fraction = decimal.Decimal('0.001')
+
+    def __init__(self):
+        self.lock = threading.Lock()
```
tarekziade: this lock can't work on a server app because we might deploy several processes behind nginx.

tarekziade: why not adding a transaction in the in-memory backend ? adding a TimeStamper like this seems to be like a special case we will never use in production so I'd rather see a transaction-based abstraction at the storage backend level

tarekziade:
Thinking about it, this solution can't work in case you get a burst of requests in the same ms: you might have requests that are *afer* that burst that have a timestamp in the past...

Natim:
> so what happens if I have 50 reqs at the same ms ? do you move timestamps in the future ? 
Yes

> you might have requests that are afer that burst that have a timestamp in the past...
No they cannot

The plan is to take the current timestamp in ms and to make sure it is greater than the previous one.

1000 changes per seconds is large enough for us and the burst will eventually ends at some point.

readinglist/utils.py
```diff
@@ -4,11 +4,16 @@
     import json  # NOQA
 
 import ast
+import time
+
 from colander import null
 
+
 # removes whitespace, newlines, and tabs from the beginning/end of a string
 strip_whitespace = lambda v: v.strip(' \t\n\r') if v is not null else v
```
tarekziade: null ?
typo, should be None

leplatrem: This is used as a *preparer* for a Colander node, which gives ``colander.null`` when no value is provided...
We could put ``not in (null, None)`` to extend its usage outside colander indeed.

tarekziade: Sorry I have not seen that null was imported - ignore my comment

leplatrem:
We'll probably have to make sure users are always sharded by server, otherwise we won't be able to guarantee revision synchronicity accross instances.

In the last commit I replaced the millisecond timestamp code by a stupid simple revision integer number, incremented by user.

If we want to restore the timestamp based approach, I guess we could also have a mix of both approaches : using revision numbers as the microseconds fraction of an epoch timestamp. This way we take advantage of having readable meaningful timestamps, without relying on accuracy magics as before.

 ``last_modified = epoch timestamp + 10⁻⁶ * (revision % 10⁶)``

*(If I'm not mistaken this would give us like 1000 revisions per millisecond per user before overlapping :)... it still would rely on sharding per user though)*

Natim:
> We'll probably have to make sure users are always sharded by server, otherwise we won't be able to guarantee revision synchronization across instances.

I don't agree there. What we tried to do in the past is to let the race being handled on the database side.

I like the approach of using timestamps in milliseconds and upgrade them in case our current timestamp is older or equal to the current one. (which is already a edge case and will not happen often)

We will definitely have less than 100 and even 1000 changes per seconds for a given record and we will display epoch in second so having twice in the same milliseconds is really really unlikely.

Also the app will refresh less than once per second so this seams to me a a bit over-engineered at that point. Also I like the idea of making sure that we are having a different date for each update of a record.

The advantage of storing timestamps in millisecond is that we have a well define and standard value.

Natim:
Replace by #63

PR #58, opened 2015-01-22
Safety check for filter values

leplatrem:


readinglist/resource.py
```diff
@@ -130,7 +130,11 @@ def _extract_filters(self, queryparams):
             if param in self.known_fields:
                 filters.append((param, value, '=='))
             if param == '_since':
-                filters.append((self.modified_field, value, '>='))
+                if isinstance(value, six.integer_types + (float,)):
```
Natim: Why would you need to handle float here?

readinglist/resource.py
```diff
@@ -130,7 +130,11 @@ def _extract_filters(self, queryparams):
...
+                    filters.append((self.modified_field, value, '>='))
```
tarekziade: I would have a set of constants for stuff like '>=' and '=='.  I have not checked but maybe Python lib has them already

readinglist/resource.py
leplatrem: timestamps were decimal values in the first version of #57

readinglist/resource.py
leplatrem: instead we could use ``operator.eq`` directly (or ``ast.Eq``)

readinglist/resource.py
Natim: So I think we don't need it anymore, right?

readinglist/resource.py
Natim: I found it more explicit without constant don't you?

readinglist/resource.py
leplatrem: Yes take it out, we'll restore it in case we change our mind in #57

readinglist/resource.py
leplatrem: the only advantage I about using constants is to have an exhaustive list somewhere.
(think backend implementation https://github.com/mozilla-services/readinglist/blob/safety-check-for-filter-values/readinglist/backend/memory.py#L55-L62)

Natim: Oh yeah.

tarekziade: The benefit with constants is avoiding typos in the string. You import and uses variables everywhere in the code.

readinglist/utils.py
```diff
@@ -18,9 +18,19 @@ def native_value(value):
...
     return type('Enum', (), enums)
+
+
+COMPARISON = Enum(
+    LT='<',
+    MIN='>=',
+    MAX='<=',
+    NOT='!=',
+    EQ='==',
+    GT='>',
```
tarekziade: That's pretty cool.   I would also move the operators dict in utils  

```python
OPERATORS = {
  COMPARISON.LT: operator.lt,
  COMPARISON.MAX: operator.le,
  COMPARISON.EQ: operator.eq,
  COMPARISON.NOT: operator.ne,
  COMPARISON.MIN: operator.ge,
  COMPARISON.GT: operator.gt,
}
```

Natim: It is what I did and we finally decided it wouldn't make sense since it is related to the memory backend only.

Natim: See https://github.com/mozilla-services/readinglist/commit/4a43c90dc215f63b8c3d6e33c9e2ce5a3ddd4a73#commitcomment-9439133

tarekziade: oh ok then

PR #60, opened 2015-01-23
Preconditions headers

leplatrem:
Based on PR #57

* [x] small refactor of single record fetch
* [x] views decorators
* [x] 304 on GET collection and GET single record
* [x] 412 basically everywhere
* [x] Format 412 errors with proper attributes

Questions: 
* Should 412 be returned on GET requests with ``If-Unmodified-Since`` ?
* ~~Group decorators into a single one ?~~

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -10,7 +10,6 @@ from pyramid.httpexceptions import HTTPTemporaryRedirect
 from pyramid_multiauth import MultiAuthenticationPolicy
 
 from readinglist import authentication
-from readinglist.resource import TimeStamp
 
 
 # Module version, as defined in PEP-0396.
@@ -61,6 +60,10 @@ def attach_http_objects(config):
     def on_new_request(event):
         # Attach objects on requests for easier access.
         event.request.db = config.registry.backend
+
+        # Track incoming request timestamp
+        event.request.timestamp = config.registry.backend.now()
+
         http_scheme = config.registry.settings.get('readinglist.http_scheme')
         if http_scheme:
             event.request.scheme = http_scheme
@@ -68,8 +71,8 @@ def attach_http_objects(config):
     config.add_subscriber(on_new_request, NewRequest)
 
     def on_new_response(event):
-        # Add timestamp info in response headers.
-        timestamp = six.text_type(TimeStamp.now())
+        # Add request timestamp info in response headers.
+        timestamp = six.text_type(event.request.timestamp)
         event.request.response.headers['Timestamp'] = timestamp.encode('utf-8')
 
         # Add backoff in response headers
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -61,8 +61,9 @@ def attach_http_objects(config):
         # Attach objects on requests for easier access.
         event.request.db = config.registry.backend
 
-        # Track incoming request timestamp
-        event.request.timestamp = config.registry.backend.now()
+        # Current revision when request comes in
+        user_id = event.request.authenticated_userid
+        event.request.revision = event.request.db.revision(user_id)
 
         http_scheme = config.registry.settings.get('readinglist.http_scheme')
         if http_scheme:
@@ -71,9 +72,9 @@ def attach_http_objects(config):
     config.add_subscriber(on_new_request, NewRequest)
 
     def on_new_response(event):
-        # Add request timestamp info in response headers.
-        timestamp = six.text_type(event.request.timestamp)
-        event.request.response.headers['Timestamp'] = timestamp.encode('utf-8')
+        # Add request revision info in response headers.
+        revision = six.text_type(getattr(event.request, 'revision', ''))
+        event.request.response.headers['Timestamp'] = revision.encode('utf-8')
 
         # Add backoff in response headers
         backoff = config.registry.settings.get("readinglist.backoff")
--- a/readinglist/backend/memory.py
+++ b/readinglist/backend/memory.py
@@ -4,6 +4,7 @@ from collections import defaultdict
 
 from readinglist.backend import BackendBase, exceptions
 from readinglist.utils import COMPARISON
+from readinglist.utils import timestamper
 
 
 tree = lambda: defaultdict(tree)
@@ -20,6 +21,9 @@ class Memory(BackendBase):
 
     def ping(self):
         return True
+
+    def now(self):
+        return timestamper.now()
 
     def create(self, resource, user_id, record):
         resource_name = classname(resource)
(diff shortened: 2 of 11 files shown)


readinglist/backend/memory.py
```diff
@@ -20,9 +21,18 @@ def flush(self):
...
+        return self._revisions[user_id]
+
+    def _bump_revision(self, user_id):
+        self._revisions[user_id] += 1
+        return self._revisions[user_id]
+
     def create(self, resource, user_id, record):
         resource_name = classname(resource)
         _id = record[resource.id_field] = self.id_generator()
+        revision = self._bump_revision(user_id)
```
tarekziade: could be a single line I guess

readinglist/backend/memory.py
```diff
@@ -35,12 +45,15 @@ def get(self, resource, user_id, record_id):
 
     def update(self, resource, user_id, record_id, record):
         resource_name = classname(resource)
+        revision = self._bump_revision(user_id)
```
tarekziade: same remark than lines 34-35

readinglist/resource.py
```diff
@@ -39,6 +42,75 @@ def wrapped_view(self, *args, **kwargs):
     return wrap
 
 
+def refresh_revision():
+    """View decorator to refresh the request revision, because it was
+    incremented during the view execution."""
+    def wrap(view):
```
tarekziade: to make traceback more meaningful, use functools.wraps here

readinglist/resource.py
```diff
@@ -39,6 +42,75 @@ def wrapped_view(self, *args, **kwargs):
...
+            return result
+        return wrapped_view
+    return wrap
+
+
+def modified_or_304():
+    """View decorator to return 304 if current revision is inferior to the one
+    specified in headers."""
+    def wrap(view):
+        def wrapped_view(self, *args, **kwargs):
```
tarekziade: functools.wraps

readinglist/resource.py
```diff
@@ -198,30 +278,30 @@ def put(self):
...
 
         updated = self.process_record(updated, old=self.record)
 
-        record = self.db.update(record_id=record_id,
+        record = self.db.update(record_id=self.record_id,
                                 record=updated,
                                 **self.db_kwargs)
         return record
 
     @resource.view(permission='readwrite')
```
tarekziade: I kinda feel we're starting to stack a lot of decorators here. The rule I usually apply for decorator usage is: can the function can work and be understood without the decorators. If not it means that you might have too much logic pushed in decorator - making the code hard to read. But OTHO if the decorator names are clear enough...

readinglist/tests/test_views_article.py
```diff
@@ -85,24 +86,129 @@ def test_string_filters_naively_by_value(self):
...
+                return MINIMALIST_ARTICLE
+
+            with mock.patch.object(self.db, 'get_all', delayed_get):
+                resp = self.app.head('/articles',
+                                     headers=self.headers)
+                revisions['fetch'] = resp.headers['Timestamp']
+
+        thread = threading.Thread(target=long_fetch)
+        thread.start()
+        resp = self.app.post_json('/articles',
```
tarekziade: if this line fails, you are leaving a thread running. use try..finally to perform the .join()

readinglist/tests/test_views_article.py
```diff
@@ -85,24 +86,129 @@ def test_string_filters_naively_by_value(self):
...
+            for i in range(100):
+                resp = self.app.post_json('/articles',
+                                          MINIMALIST_ARTICLE,
+                                          headers=self.headers)
+                current = int(resp.json['last_modified'])
+                obtained.append(current)
+
+        thread1 = threading.Thread(target=hit_post)
+        thread2 = threading.Thread(target=hit_post)
+        thread1.start()
```
tarekziade: it might be wise to cretate a _create_thread function in your test, so you can make sure all threads are cleaned up on tearDown

Natim:
As I explained in another comment, I would prefer to keep up with Timestamp because Revision adds complexity for the end user as well as for the developer.

readinglist/__init__.py
```diff
@@ -61,6 +60,11 @@ def attach_http_objects(config):
     def on_new_request(event):
         # Attach objects on requests for easier access.
         event.request.db = config.registry.backend
+
+        # Current revision when request comes in
+        user_id = event.request.authenticated_userid
+        event.request.revision = event.request.db.revision(user_id)
```
Natim: How can you obtain a global revision for a given user? Shouldn't the revision be linked to the article?

readinglist/backend/memory.py
```diff
@@ -13,6 +13,7 @@ class Memory(BackendBase):
     def __init__(self, *args, **kwargs):
         super(Memory, self).__init__(*args, **kwargs)
         self._store = tree()
+        self._revisions = defaultdict(int)
```
Natim: This means the memory backend is not usable in a scalable environment. We should probably tell this in the documentation.

leplatrem: It has never been its purpose though :)

readinglist/__init__.py
leplatrem: It may be a problem of vocabulary. Here revision was meant to replace the timestamp notion.

*(I didn't want to manipulate revisions in timestamps variables.)*

We should probably leave timestamp opaque, and use timestamp wording everywhere, to avoid such confusion.

readinglist/resource.py
leplatrem: We could rename some of them:
* ``exists_or_404`` --> ``fetch_record_or_404``
* ``refresh_revision`` --> move the logics elsewhere ? like ``if request.method in ('POST', 'PUT', 'PATCH', 'DELETE')`` in NewResponse event ?

readinglist/tests/test_views_article.py
leplatrem: good catch thanks

readinglist/backend/memory.py
Natim: Yes :)

readinglist/__init__.py
ametaireau: Yes, timestamp seems clearer (at least to me)

readinglist/backend/memory.py
ametaireau: I prefer to keep these lines separated, it makes the reading easier to follow.

readinglist/resource.py
ametaireau: It also seems we're using decorators to add validation logic. We probably can find other means to add the same validation logic. With pyramid / cornice, it usually passes trough attributes.

PR #63, opened 2015-01-27
Rework notion of revision

leplatrem:
* Renamed revision to timestamp
* Base timestamp on current millisecond
* Slide into the future when a time collision occurs
* Put millisecond timestamps everywhere (eg. ``stored_on``)

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -61,9 +61,9 @@ def attach_http_objects(config):
         # Attach objects on requests for easier access.
         event.request.db = config.registry.backend
 
-        # Current revision when request comes in
+        # Current timestamp when request comes in
         user_id = event.request.authenticated_userid
-        event.request.revision = event.request.db.revision(user_id)
+        event.request.timestamp = event.request.db.timestamp(user_id)
 
         http_scheme = config.registry.settings.get('readinglist.http_scheme')
         if http_scheme:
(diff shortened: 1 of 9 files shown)


readinglist/__init__.py
```diff
@@ -72,9 +72,9 @@ def on_new_request(event):
     config.add_subscriber(on_new_request, NewRequest)
 
     def on_new_response(event):
-        # Add request revision info in response headers.
-        revision = six.text_type(getattr(event.request, 'revision', ''))
-        event.request.response.headers['Timestamp'] = revision.encode('utf-8')
+        # Add request timestamp info in response headers.
+        timestamp = six.text_type(getattr(event.request, 'timestamp', ''))
```
Natim: I don't want to return an empty header in case we don't have any timestamp, we should return the server timestamp.
Also I wonder if we don't want to return the real server Timestamp in Timestamp and the Last-Modified header with the current collection timestamp.

Last point is that I think this db.timestamp thing should be related to the user collection and not only the user, as soon as we will have to manage more than one collection we will have problems.

readinglist/backend/memory.py
```diff
@@ -22,18 +22,30 @@ def flush(self):
...
+
+        .. note ::
+
+            Here it is assumed that if requests from the same user burst in,
+            the time will slide into the future. It is not problematic since
+            the timestamp notion is opaque, and behaves like a revision number.
+        """
+        previous = self._timestamps.get(user_id)
+        current = utils.msec_time()
+        if previous and previous == current:
```
Natim: ``previous <= current``

readinglist/backend/memory.py
```diff
@@ -22,18 +22,30 @@ def flush(self):
...
+            current += 1
```
Natim: previous += 1

leplatrem:
@Natim r?

readinglist/resource.py
```diff
@@ -128,6 +115,13 @@ def __init__(self, request):
         self.record = None
         self.known_fields = [c.name for c in self.mapping.children]
 
+    def add_timestamp_header(self):
+        """Add current timestamp in response headers, when request comes in.
+        """
+        timestamp = self.db.timestamp(**self.db_kwargs)
+        timestamp = six.text_type(timestamp).encode('utf-8')
+        self.request.response.headers['Timestamp'] = timestamp
```
Natim: Didn't you want to rename it Last-Modified?

Issue #40, opened 2015-01-19
Implement preconditions behaviour


leplatrem:
* 304 responses when if-modified (*facultative*)
* 412 when if-unmodified (*facultative*)

PR #64, opened 2015-01-27
Add timestamp in 304 and 412 responses (ref #40)

leplatrem:


Natim:
Why would you like to remove Content-Type? This is important information for both json type and charset.

leplatrem:
I was removing the default one (html/utf8) because it is set by Cornice and our views anyway...

I just reworked the way I pass the last-modified header in error views.

@Natim r?

PR #65, opened 2015-01-28
Log incoming requests to the console.

Natim:
That would ease debugging the client a little. Thanks!

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -1,6 +1,9 @@
+# -*- coding: utf-8 -*-
 """Main entry point
 """
+from __future__ import print_function
 import pkg_resources
+from datetime import datetime
 
 from pyramid.config import Configurator
 from pyramid.events import NewRequest, NewResponse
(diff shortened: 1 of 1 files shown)


Natim:
15 pour @tarekziade qui avait prédit que dès qu'on mettrait un middleware ça ferait péter les tests.

n1k0:
Let me apply the patch and tell.

n1k0:
Got `pkg_resources.DistributionNotFound: Paste` on a make serve.

Natim:
J'ai ajouté paste dans les requirements mais ton setup.py à pas dû être redeploy.

    .venv/bin/pip install -e ./

leplatrem:
Even if translogger seems advocated in Pyramid docs http://docs.pylonsproject.org/docs/pyramid/en/latest/narr/logging.html#request-logging-with-paste-s-translogger, I find it quite overkill to introduce Paste just for that :( Moreover, it's not p3.4 compatible!

Natim:
Ok let's find another solution then.

leplatrem:
Appart from nitpicks, LGTM :)

Natim:
Updated with your comments.

PR #66, opened 2015-01-28
[WIP] Refactor tests of resource views

leplatrem:
Rationale:

* Switch from functional to unit test for resource views
* Bypass Pyramid machinery to test resource views
* Do not rely on Article inherited views to test resource

WIP:

* Unstable tests for 304 and 412 errors
* Record validation tests
* JSON errors tests
* Move sorting test from article to resource
* Move filter test from article to resource
* Move timestamps from article to resource


leplatrem:
```
----------------------------------------------------------------------
Ran 160 tests in 0.477s

OK
```

leplatrem:
Now requires #74 

One test failing only: Looks like cornice CatchErrors has no effect :(

readinglist/views/errors.py
```diff
@@ -101,3 +102,8 @@ def error(context, request):
         content_type='application/json')
 
     return reapply_cors(request, response)
+
+
+@view_config(context=HTTPBadRequest)
+def json_error(context, request):
+    return context
```
leplatrem: @Natim @tarekziade @ametaireau I had to add this just for the [test app that I setup here](https://github.com/mozilla-services/readinglist/pull/66/files#diff-c445a1cd6e41331ed421066d9a202915R43)

*(it is not necessary when executing the views via httppie or webtest)*

     Note that using ``cornice.tests.support.CatchErrors`` had no effect.
Any hint ?

PR #67, opened 2015-01-29
Patch Service to allow cors.

Natim:
@n1k0 If you want to try this.


n1k0:
@natim works great!

r=me with simple tests added.

PR #72, opened 2015-01-31
mark as read doesn't handle the marked_read_by parameter.

Natim:
I tried to mark my article as read using:

    http PATCH https://readinglist.dev.mozaws.net/v0/articles/132cdcbc40374097834134c8cc8b1b8a \ 
        Authorization:"Bearer <token>" \
        marked_read_by=Inspiron marked_read_on=1422691406308 unread=False

But as a matter of fact the marked_read_by parameter is not taken into account.

### DIFF
--- a/readinglist/tests/test_views_article.py
+++ b/readinglist/tests/test_views_article.py
@@ -62,6 +62,21 @@ class ArticleModificationTest(BaseWebTest, unittest.TestCase):
         self.assertIsNone(resp.json['marked_read_by'])
         self.assertIsNone(resp.json['marked_read_on'])
 
+    def test_patch_marked_read_are_taken_into_account(self):
+        mark_read = {
+            'unread': False,
+            'marked_read_by': 'FxOS',
+            'marked_read_on': 1234}
+        resp = self.app.patch_json(self.url, mark_read, headers=self.headers)
+        body = resp.json
+        self.assertEqual(body['marked_read_by'], mark_read['marked_read_by'])
+        self.assertEqual(body['marked_read_on'], mark_read['marked_read_on'])
+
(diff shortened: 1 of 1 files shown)


Natim:
Ok I tracked down this bug and it is a red hearing.

readinglist/tests/test_views_article.py
```diff
@@ -53,8 +53,14 @@ def test_mark_by_and_on_are_set_to_none_if_unread_is_true(self):
...
-        self.assertIsNotNone(resp.json['marked_read_by'])
-        self.assertIsNotNone(resp.json['marked_read_on'])
+        body = resp.json
+        self.assertEqual(body['marked_read_by'], mark_read['marked_read_by'])
+        self.assertEqual(body['marked_read_on'], mark_read['marked_read_on'])
+
+        resp = self.app.get(self.url, headers=self.headers)
+        body = resp.json
+        self.assertEqual(body['marked_read_by'], mark_read['marked_read_by'])
+        self.assertEqual(body['marked_read_on'], mark_read['marked_read_on'])
```
leplatrem: I don't understand what you are trying to do here...

The test says *"mark_by_and_on_are_set_to_none"* and the what the test currently asserts is "it is not none before", then change unread, then "it is none after".

If you need to assert other things, then start a new test method with an explicit spec :)

Natim: I wanted to check that the ``marked_read_by`` is correctly set to the value provided with the request.

leplatrem: you could then move this to another test called "patch marked by and on are taken into account"

Natim:
Updated.

Issue #19, opened 2014-12-22
Versionned protocol


leplatrem:
* Discussion we had for loop https://bugzilla.mozilla.org/show_bug.cgi?id=1025792
* And for fxA https://mail.mozilla.org/pipermail/sync-dev/2013-September/000486.html

Natim:
Described here: http://readinglist.readthedocs.org/en/latest/versionning.html#versionning
Implemented here: https://github.com/mozilla-services/readinglist/blob/master/readinglist/__init__.py#L43

Issue #18, opened 2014-12-22
Deprecation indicators


leplatrem:

ametaireau:
Something else I think we could support is parameter renames.

For instance, for loop we had to change some parameter names at some point. The only way we had to deal with that was to accept both old and new parameters on the server side. That's fine, but then the client don't notice anything.

I would like to have a deprecation channel also report any chages like that. I'm not sure how this should be implemented explicitely, but headers seems a good way to convey such information.

ametaireau:
@rfk in the bug you referred to, you're mentioning "513 GO AWAY", but I cannot find this in Any specification.

As this sounds rather a client error rather than a client one, I feel it should be a 4XX error code.
After a quick review of what's available, I found the `426 / Upgrade Required` one (RFC 2817) which is for TLS upgrade normally.

Maybe we should go for this?

rnewman:
I'd like to keep this narrowly focused on protocol/service-level deprecation — that is, telling a client that it needs to upgrade, or that it needs to tell the user that the service is going away. That's because I don't think we'd build anything more fine-grained than that in all of our clients.

E.g., wrt renames, I can't imagine the client being able to do anything intelligent here above and beyond turning itself off until it upgrades, so I suspect that's how we'd do it: shift to the next version of the protocol, wait a few releases, then have the old protocol endpoint start sending deprecation indicators to clients.

I think "513" was a typo of 503:

```
Hard EOL:  all requests return a 503, with an X-Weave-Alert header that includes a JSON object with code:hard-eol, like this:

    513 GO AWAY
```
As to client error versus server error: I think this is the server refusing to handle a request, rather than the client creating a malformed request.

We found 426, but IIRC it wasn't really suitable — yes, it's for TLS (and thus will misbehave when proxies get in the way, and we probably shouldn't reuse it for a different purpose), but it also means "I can't continue this conversation until you upgrade, and when you do, this URL is still good", while for our purposes we want "I understood you, but I'm refusing to act, and you need to go to a different URL after you upgrade". To me that sounds like a "justified 503", or similar.

rfk:
> you're mentioning "513 GO AWAY", but I cannot find this in Any specification.

Indeed; turns out we made up a new code because none of the existing ones seemed to fit: https://bugzilla.mozilla.org/show_bug.cgi?id=895518#c11

rfk:
[...]

PR #73, opened 2015-02-02
Setup readinglist documentation. (refs #18, refs #19)

Natim:
Is it enough for you or do you want to spawn a virtualenv for the documentation dependencies?


Natim:
R? @leplatrem

leplatrem:
What do you think if we only put what's already implemented ?

When we'll implement a new feature, we'll copy-paste the piece of doc from the Wiki and put it here in the docs.

This way, there is no confusion between *planned* and *available* features. Plus, we can also make sure the implementation matches the documentation during pull request review

docs/index.rst
```diff
@@ -0,0 +1,32 @@
+.. Reading List documentation master file, created by
+   sphinx-quickstart on Mon Feb  2 15:08:06 2015.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to Reading List's documentation!
```
leplatrem: Reading List documentation

Usually you would use the apostrophe when it means "possession" (like Remy's flat)

docs/index.rst
```diff
@@ -0,0 +1,32 @@
...
+========================================
+
+*Reading list* is a service to keep track of a list of articles to read.
+The state of the reading list is synchronized between devices.
```
leplatrem: This is a service attached to Firefox Accounts that enable users to keep their Reading List in sync across all of their devices.

source: https://wiki.mozilla.org/Mobile/FxA_-_Reading_List

docs/index.rst
Natim: Well this is made up by Sphinx quickstart :)

Natim: Plus it looks right according to: http://english.stackexchange.com/a/6736

docs/Makefile
```diff
@@ -0,0 +1,180 @@
+# Makefile for Sphinx documentation
```
ametaireau: I always feel this Makefile is kinda superfluous: we probably should just do the call to sphinx-build and that would be easier: we don't use the other targets.

Natim: I like to keep the Sphinx default Makefile for all projects. It helps people coming to the project to feel home.

docs/api.rst
```diff
@@ -0,0 +1,537 @@
...
+GET /
+-----
+
+The returned value is a JSON mapping containing:
+
+- ``hello``: the name of the service (``"reading list"``)
+- ``version``: complete version (``"X.Y.Z"``)
+- ``url``: absolute URI (without a trailing slash) of the API (*can be used by client to build URIs*)
+- ``eos``: date of end of support in ISO 8601 format (``"yyyy-mm-dd"``, null if unknown)
+- ``documentation``: The url to the service documentation
```
ametaireau: we should add "(here !)"

docs/api.rst
```diff
@@ -0,0 +1,537 @@
...
+
+
+GET /__heartbeat__
+------------------
+
+Return the status of each service the *reading list* depends on. The returned value is a JSON mapping containing:
+
+- ``database`` true if operational
+
+Return ``200`` if the connection with each service is working properly and ``503`` if something doesn't work.
```
ametaireau: Hmm, we should wrap at 80 chars :)

docs/api.rst
```diff
@@ -0,0 +1,537 @@
+#################
+API Documentation
+#################
+
+
+GET /
+-----
```
ametaireau: We probably should use `====` for first level titles, and `----` for second level ones.

docs/index.rst
ametaireau: Anyway, "welcome to" seems weird to me. I believe we should remove the "welcome" part.

docs/conf.py
```diff
@@ -0,0 +1,268 @@
...
+#texinfo_appendices = []
+
+# If false, no module index is generated.
+#texinfo_domain_indices = True
+
+# How to display URL addresses: 'footnote', 'no', or 'inline'.
+#texinfo_show_urls = 'footnote'
+
+# If true, do not generate a @detailmenu in the "Top" node's menu.
+#texinfo_no_detailmenu = False
```
ametaireau: Wow, this is big.

docs/backoff.rst
```diff
@@ -0,0 +1,36 @@
+##################
+Backoff indicators
+##################
+
+Backoff header on heavy load
```
ametaireau: We probably should add anchors here so that we can link there from other docs easily.

Natim:
Updated.

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+==================
+
+Return the status of each service the *reading list* depends on. The
+returned value is a JSON mapping containing:
+
+- ``database`` true if operational
+
+Return ``200`` if the connection with each service is working properly
+and ``503`` if something doesn't work.
+
```
leplatrem: What do you think if we put these two *generic* endpoints in a separate doc file ?

docs/authentication.rst
```diff
@@ -0,0 +1,38 @@
...
+
+**Reading list scope**
+
+The *reading list* API will eventually have to handle a dedicated OAuth scope (e.g.
+``readinglist``, ``readinglist:read``, ``readinglist:write``). This will help users
+to delegate access to the readinglist to third party apps
+
+So far the FxA server only handles the ``profile`` scope.
+
+See https://github.com/mozilla-services/readinglist/issues/16.
```
leplatrem: We could mention the BasicAuth configuration for local testing or during development

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+If the request header ``If-Unmodified-Since`` is provided, and if the record has
+changed meanwhile, a ``412 Precondition failed`` error is returned.
+
+:note:
+    Once deleted, an article will appear in the collection with a deleted status
+    (``status=2``) and will have most of its fields empty.
+
+:note:
+    The server will have to implement an internal mechanism to will keep track of deleted items,
+    and purge them eventually.
```
leplatrem: Since this is not implemented, should we mention it here ?

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+    Changing ``read_position`` never generates conflicts.
+
+:note:
+    ``read_position`` can only be changed for a greater value than the current one.
+
+:note:
+    If ``unread`` is changed to false, ``marked_read_on`` and ``marked_read_by`` are expected to be provided.
+
+:note:
+    If ``unread`` was already false, ``marked_read_on`` and ``marked_read_by`` are not updated with provided values.
```
leplatrem: Since these 3 details are not implemented, should we mention them here ?

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+
+:note:
+    Deleted items should be taken into account for URL unicity.
+
+If an article is created with an URL that already exists, a ``303 See Other`` response
+is returned to indicate the existing record.
+
+The response body is a JSON mapping, with the following attribute:
+
+- ``id``: the id of the conflicting record
```
leplatrem: Not implemented either

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+If the posted values are invalid (e.g. *added_on is not an integer*) an error response is returned with status ``400``. `See details on error responses <https://github.com/mozilla-services/readinglist/wiki/API-Design-proposal#error-responses>`_.
+
+
+:note:
+    The ``status`` can take only ``0`` (OK) and ``1`` (archived), even though
+    the server sets it to ``2`` when including deleted articles in the collection.
+
+:note:
+    *(undecided)* For some cases, it can make sense for the server to fix arbitrarily
+    validation errors on records (e.g. truncating long titles).
```
leplatrem: not decided

docs/api.rst
```diff
@@ -0,0 +1,339 @@
...
+
+- ``<prefix?><attribute name>``: filter by value(s)
+- ``_since``: polling changes
+- ``_sort``: order list
+- ``_limit``: pagination max size
+
+Some additional internal parameters are used by pagination. Client should not
+be aware of them, since they are set and provided through the ``Next-Page`` header.
+
+- ``_page_token``: pagination continuation token
```
leplatrem: Not decided / implemented

Natim:
Updated.

PR #77, opened 2015-02-03
Conditional operations on read_position (ref #6)

leplatrem:
* [x] read position is defaulted to 0
* [x] read position is ignored if inferior to current
* [x] read position is reset to 0 when article marked unread
* [x] require marked read by|on when marking as read
* [x] ignore marked read by|on if already read
* [x] do not update last_modified if fields have not changed
* [x] do not update collection last modified if fields have not changed
* [x] update docs


leplatrem:
@Natim @ametaireau r?

docs/api.rst
```diff
@@ -280,23 +281,26 @@ If the request header ``If-Unmodified-Since`` is provided, and if the record has
...
-       If ``unread`` was already false, ``marked_read_on`` and ``marked_read_by`` are not updated with provided values.
+:note:
+    If ``unread`` was already false, ``marked_read_on`` and ``marked_read_by``
+    are not updated with provided values.
 
 :note:
-    If ``unread`` is changed to true, ``marked_read_by`` and ``marked_read_on``
-    are changed automatically to null.
+    If ``unread`` is changed to true, ``marked_read_by``, ``marked_read_on``
+    and ``read_position`` are resetted to their default value.
```
ametaireau: [Apparently](http://forum.wordreference.com/showthread.php?t=70776), "resetted" doesn't exist.

ametaireau: (and should be replaced by "reset")

readinglist/resource.py
```diff
@@ -312,7 +314,12 @@ def put(self):
             existing = None
 
         new_record = self.request.validated
-        new_record[self.id_field] = record_id
+
+        new_id = new_record.setdefault(self.id_field, record_id)
+        if new_id != record_id:
+            error_msg = 'Record id do not match'
```
ametaireau: Do not match existing one?

readinglist/views/article.py
```diff
@@ -41,12 +41,15 @@ class ArticleSchema(ResourceSchema):
...
-    status = SchemaNode(colander.Integer(), missing=0)
+    status = SchemaNode(colander.Integer(), missing=0,
+                        validator=colander.Range(min=0, max=1))
     favorite = SchemaNode(colander.Boolean(), missing=False)
     unread = SchemaNode(colander.Boolean(), missing=True)
     is_article = SchemaNode(colander.Boolean(), missing=True)
     excerpt = SchemaNode(String(), missing="")
 
+    read_position = SchemaNode(colander.Integer(), missing=0,
+                               validator=colander.Range(min=0))
```
ametaireau: we don't have any max number of words (even theoretical?) I believe we should :)

ametaireau:
That looks good to me. I added a few remarks.

readinglist/views/article.py
leplatrem: I don't think so, this is something that could be discussed in #3 . So far I consider read_position as an abstract notion

ametaireau: ok

PR #78, opened 2015-02-03
Create userid based on userid hmac.

Natim:
- username:password for Basic Auth
 - profile user id for Oauth Bearer token

It is important to hmac the Oauth userid in order to keep user privacy in between Oauth servers.
It is important to patch the Basic auth backend to prevent users data to get access with any password.

### DIFF
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -1,3 +1,5 @@
+import hashlib
+import hmac
 from zope.interface import implementer
 
 from pyramid import authentication as base_auth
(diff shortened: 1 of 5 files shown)


readinglist/authentication.py
```diff
@@ -18,7 +20,15 @@ def check_credentials(username, password, request):
     """
     settings = request.registry.settings
     is_enabled = settings.get('readinglist.basic_auth_backdoor')
-    return ["basicauth_%s" % username] if is_enabled else []
+
+    hmac_key = request.registry.settings.get(
```
ametaireau: should we name this `hmac_secret`?

ametaireau: Also, we probably want to read this from the `__init__` and pass it as an argument rather than reading it from the authentication module.

Natim: We should probably refactor all settings access then.

Natim: I have created #81 about this.

Natim:
@ametaireau r?

PR #79, opened 2015-02-03
Oauth flow improvments

Natim:
- [x] Improve the backend to accept dummy resources.
- [ ] Store the user state on the server side
- [ ] Store the redirect URI on the server side


readinglist/backend/simpleredis.py
```diff
@@ -39,7 +39,7 @@ def ping(self):
 
     def last_collection_timestamp(self, resource, user_id):
         """Return the last timestamp for the resource collection of the user"""
-        resource_name = classname(resource)
+        resource_name = resource_id(resource)
```
ametaireau: We should decide between `resource_id` and `resource_name` here :)

leplatrem: +1 for resource_id everywhere :)

readinglist/tests/test_utils.py
```diff
@@ -36,3 +36,13 @@ def test_does_remove_middle_spaces(self):
 
     def test_idempotent_for_null_values(self):
         self.assertEqual(strip_whitespace(colander.null), colander.null)
+
+
+class ResourceIdTest(unittest.TestCase):
+    def test_string_resource_id(self):
+        self.assertEqual(resource_id('toto'), 'toto')
```
leplatrem: I'm not sure about this dummy "string" resource.

I was thinking about "field unicity" implementation: the backend needs a handle on the schema/resource to obtain some details about fields (types, contraints, ...) to initialize the "storage constraints"

Let's think about it

leplatrem: I thought that the resource id might be a property/method of the resource class instead.

For example, if we imagine "dynamic resource of cloud storage" : 

```python
@crud(collection_path='/{resource_id}',
      item_path='/{resource_id}/{id}')
class DynamicResource(...):
 ....
    def __init__(...):
        self.resource_id = self.request.match_dict['resource_id']
```

PR #87, opened 2015-02-04
Do not compute userid hmac if BasicAuth is not enabled

leplatrem:


Natim:
You should also make sure username is not empty.

ametaireau:
LGTM

PR #91, opened 2015-02-04
Rename _id with id.

Natim:


leplatrem:
> <natim> oui c'est une erreur qu'il s'appelle _id

Having ``_id`` instead of ``id`` was not a mistake, but a way to mimic *CouchDB*, *ElasticSearch*, *MongoDB*, *Python Eve*, *Meteor*, ...

Initially I even had in mind to rename ``last_modified`` to ``_last_modified``, in order to distinguish internal fields from domain specific fields... 

Maybe add some ``_meta`` for HATEOA eventually !

I don't want to interfere to much here, but just so that you know : ``_id`` was a choice that once seemed reasonable to me :)

Maybe we could open a vote ? :)

n1k0:
I'm fine either way (sorry guys :p)

Natim:
Well it is not what we defined in the protocol here: https://github.com/mozilla-services/readinglist/wiki/API-Design-proposal#data-model

What is the purpose of having this ``_`` prefix? Telling the user that it is automatic fields? I don't think it is necessary.

leplatrem:
@Natim I kind of agree :) ``id`` would just be fine. Especially if @rnewman agreed on the API proposal :)
That will imply breaking the clients code, but as you once said, we shouldn't care about retrocompatibility at this phase of the project !

PR #92, opened 2015-02-04
Disable PUT endpoint for articles (fixes #42)

### LINKED ISSUE #42: Disable the PUT endpoint
Even though it makes sense in a RESTful approach, it is wise to disable it since it's not relevant in the context of readinglist

leplatrem:


leplatrem:
@ametaireau r?

Issue #35, opened 2015-01-19
Batch endpoint payload optimizations


leplatrem:
Current approach for BATCH operations is to provide a series of requests and return a list of responses.

See https://github.com/mozilla-services/readinglist/wiki/API-Design-proposal#post-batch

This can result in a pretty massive response payload, especially while importing hundreds (thousands?) of articles.

The puporse of this issue is to brainstorm around a sound formalism that would allow to reduce response payloads.

For example, we can imagine adding a request header ``Light-Response`` or a querystring parameter to reduce the response JSON layout/content.
https://github.com/mozilla-services/readinglist/wiki/API-Design-proposal#massive-operations

leplatrem:
Do you mind if I reopen this?
It was targetting response payload optimization, which was not covered by #2...

Natim:
Oh yes my bad.

PR #93, opened 2015-02-05
Implement batch api

leplatrem:
* [x] update docs
* [x] validation schema of batch 
* [x] invoke subrequests http://docs.pylonsproject.org/docs/pyramid/en/latest/narr/subrequest.html
* [x] catch errors
* [x] serialize responses
* [x] merge headers specified with current request
* [x] defaults values for requests
* [x] limit number of operations (from settings ?)

### DIFF
--- a/docs/batch.rst
+++ b/docs/batch.rst
@@ -1,0 +1,118 @@
+################
+Batch operations
+################
+
+.. _batch:
+
+POST /batch
+===========
+
+**Requires an FxA OAuth authentication**
+
+The POST body is a mapping, with the following attributes:
+
+- ``requests``: the list of requests
+- ``defaults``: (*optional*) values in common for all requests
+
+ Each request is a JSON mapping, with the following attribute:
+
+- ``method``: HTTP verb
+- ``path``: URI
+- ``body``: a mapping
+- ``headers``: (*optional*), otherwise take those of batch request
+
+::
+
+    {
+      "defaults": {
+        "method" : "POST",
+        "path" : "/articles",
+        "headers" : {
+          ...
+        }
+      },
+      "requests": [
+        {
+          "body" : {
+            "title": "MoFo",
+            "url" : "http://mozilla.org"
+          }
+        },
+        {
+          "body" : {
+            "title": "MoCo",
+            "url" : "http://mozilla.com"
+          }
+        },
+        {
+          "method" : "PATCH",
+          "path" : "/articles/409",
+          "body" : {
+            "read_position" : 3477
+          }
+        }
+      ]
+    ]
+
+
+The response body is a list of all responses:
+
+::
+
+    {
+      "defaults": {
+        "path" : "/articles",
+      },
+      "responses": [
+        {
+          "path" : "/articles/409",
+          "status": 200,
+          "body" : {
+            "id": 409,
+            "url": "...",
+            ...
+            "read_position" : 3477
+          },
+          "headers": {
+            ...
+          }
+        },
+        {
+          "status": 201,
+          "body" : {
+            "id": 411,
+            "title": "MoFo",
+            "url" : "http://mozilla.org",
+            ...
+          },
+        },
+        {
+          "status": 201,
+          "body" : {
+            "id": 412,
+            "title": "MoCo",
+            "url" : "http://mozilla.com",
+            ...
+          },
+        },
+      ]
+    ]
+
+
+:note:
+
+     The responses are not necessarily in the same order of the requests.
+
+
+Pros & Cons
+:::::::::::
+
+* This respects REST principles
+* This is easy for the client to handle, since it just has to pile up HTTP requests while offline
+* It looks to be a convention for several REST APIs (`Neo4J <http://neo4j.com/docs/milestone/rest-api-batch-ops.html>`_, `Facebook <https://developers.facebook.com/docs/graph-api/making-multiple-requests>`_, `Parse <ttps://parse.com/docs/rest#objects-batch>`_)
+* Payload of response can be heavy, especially while importing huge collections
+* Payload of response must all be iterated to look-up errors
+
+:note:
+
+    A form of payload optimization for massive operations is planned.
--- a/docs/batch.rst
+++ b/docs/batch.rst
@@ -60,9 +60,6 @@ The response body is a list of all responses:
 ::
 
     {
-      "defaults": {
-        "path" : "/articles",
-      },
       "responses": [
         {
           "path" : "/articles/409",
@@ -79,6 +76,7 @@ The response body is a list of all responses:
         },
         {
           "status": 201,
+          "path" : "/articles",
           "body" : {
             "id": 411,
             "title": "MoFo",
@@ -88,6 +86,7 @@ The response body is a list of all responses:
         },
         {
           "status": 201,
+          "path" : "/articles",
           "body" : {
             "id": 412,
             "title": "MoCo",
@@ -99,9 +98,14 @@ The response body is a list of all responses:
     ]
 
 
+:warning:
+
+    Since the requests bodies are necessarily mappings, posting arbitrary data
+    (*like raw text or binary*)is not supported.
+
 :note:
 
-     The responses are not necessarily in the same order of the requests.
+     The responses are in the same order of the requests.
 
 
 Pros & Cons
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -87,7 +87,6 @@ class AuthorizationPolicy(object):
         Authenticated users only are allowed.
         """
         PERMISSIONS = {
-            'batch': Authenticated,
             'readonly': Authenticated,
             'readwrite': Authenticated,
         }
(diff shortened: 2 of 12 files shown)


leplatrem:
@Natim @tarekziade @ametaireau This is a first shot on the implementation of ``POST /batch``. If you see something outrageously wrong, please speak up! :)

docs/batch.rst
```diff
@@ -0,0 +1,118 @@
...
+        "headers" : {
+          ...
+        }
+      },
+      "requests": [
+        {
+          "body" : {
+            "title": "MoFo",
+            "url" : "http://mozilla.org"
+          }
```
leplatrem: Note: update this to put body as string

Natim: Is this a requirements? It seems strange to me to have JSON encoded string inside a JSON. I rather prefer to have a nested object instead.

docs/batch.rst
```diff
@@ -0,0 +1,118 @@
...
+      ]
+    ]
+
+
+The response body is a list of all responses:
+
+::
+
+    {
+      "defaults": {
```
Natim: I think we should not have defaults for the response. Also it makes sense for the request.

docs/batch.rst
```diff
@@ -0,0 +1,118 @@
...
+            ...
+          },
+        },
+      ]
+    ]
+
+
+:note:
+
+     The responses are not necessarily in the same order of the requests.
```
Natim: Well it should. I doesn't make sense to have them in another order since they don't have ids how to you merge the results if there are not in the same order?

readinglist/authentication.py
```diff
@@ -87,6 +87,7 @@ def permits(self, context, principals, permission):
         Authenticated users only are allowed.
         """
         PERMISSIONS = {
+            'batch': Authenticated,
```
Natim: Why do we need a batch permission? Are all verbs authorized for the batch endpoint or only PATCH on the resource?

docs/batch.rst
leplatrem: Yes, you're right. But we should support both then *(suppose we want to support xml or form encoded payloads)*

docs/batch.rst
leplatrem: +1, plus it's kind of tricky to implement :)

docs/batch.rst
Natim: Well we don't :)

docs/batch.rst
leplatrem: We'll think of something neat for #35

readinglist/authentication.py
Natim: Should we rather use the endpoint permission returning 401 if needed?

readinglist/tests/resource/__init__.py
```diff
@@ -16,13 +12,8 @@ def setUp(self):
         self.resource = BaseResource(self.get_request())
 
     def get_request(self):
-        request = mock.MagicMock(headers={})
+        request = DummyRequest()
```
Natim: Miam

readinglist/views/batch.py
```diff
@@ -0,0 +1,115 @@
...
+    path = colander.SchemaNode(colander.String(),
+                               validator=colander.Regex('^/'))
+    headers = colander.SchemaNode(colander.Mapping(unknown='preserve'),
+                                  validator=string_values,
+                                  missing=colander.drop)
+    body = colander.SchemaNode(colander.String(),
+                               missing=colander.drop)
+
+
+class BatchPayloadSchema(colander.MappingSchema):
```
Natim: Comment est géré le default là dessus ?

docs/batch.rst
leplatrem: I prefer to leave this, in case batch requests are executed asynchronously. I can add a counter then :)

Natim: Asynchronously doesn't means that the results should be shuffled.

readinglist/authentication.py
leplatrem: Indeed, why not leave this endpoint open and rely on sub-requests permission checks...

readinglist/views/batch.py
leplatrem: not yet :)

readinglist/authentication.py
Natim: Yes that's my point.

readinglist/views/batch.py
Natim: Ok I just saw the checkbox in the PR description.

leplatrem:
@Natim r?

docs/batch.rst
Natim: Should we remove it from the docs since we don't implement it.

docs/batch.rst
```diff
@@ -0,0 +1,122 @@
...
+ Each request is a JSON mapping, with the following attribute:
+
+- ``method``: HTTP verb
+- ``path``: URI
+- ``body``: a mapping
+- ``headers``: (*optional*), otherwise take those of batch request
+
+::
+
+    {
```
ametaireau: If we provide examples, we should do it with HTTPie so that it's easier for clients to reproduce.

docs/batch.rst
ametaireau: Yeah, I concur here: we currently don't and in case we want to, we can re-assess this formalism to include arbitrary strings.

docs/batch.rst
```diff
@@ -0,0 +1,122 @@
...
+
+
+:warning:
+
+    Since the requests bodies are necessarily mappings, posting arbitrary data
+    (*like raw text or binary*)is not supported.
+
+:note:
+
+     The responses are in the same order of the requests.
```
ametaireau: I don't believe this is english enough :)

`Responses are provided in the same order than requests`

readinglist/views/batch.py
```diff
@@ -0,0 +1,165 @@
+import json
+
+import colander
```
ametaireau: `import x` should go after `from x import y`

docs/batch.rst
```diff
@@ -0,0 +1,122 @@
...
+
+POST /batch
+===========
+
+**Requires an FxA OAuth authentication**
+
+The POST body is a mapping, with the following attributes:
+
+- ``requests``: the list of requests
+- ``defaults``: (*optional*) values in common for all requests
```
ametaireau: While reading the code (without more information), I found it hard to unsderstand what this was for. We probably should rename this argument to `default_values` or `default_request_values`, that would make it clearer.

readinglist/views/batch.py
```diff
@@ -0,0 +1,165 @@
...
+    body = colander.SchemaNode(colander.Mapping(unknown='preserve'),
+                               missing=colander.drop)
+
+
+class BatchPayloadSchema(colander.MappingSchema):
+    requests = colander.SchemaNode(colander.Sequence(),
+                                   BatchRequestSchema())
+
+    def deserialize(self, cstruct=colander.null):
+        """Preprocess received data
```
ametaireau: If this is one line, should we end the `"""` on the same line?

readinglist/views/batch.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        if cstruct is colander.null:
+            return colander.null
+
+        # See if defaults was specified in payload
```
ametaireau: nit: add a dot at the end of the sentence.

ametaireau: At least, we should rephrase the comment here: "See if default arguments were specified"

readinglist/views/batch.py
```diff
@@ -0,0 +1,165 @@
...
+        defaults = cstruct.get('defaults', {})
+        if isinstance(defaults, dict):
+            defaults.setdefault('path', '/')
+        defaults = BatchRequestSchema().deserialize(defaults)
+
+        # Fill requests values with defaults
```
ametaireau: nit: dot at the end of the sentence.

readinglist/views/batch.py
```diff
@@ -0,0 +1,165 @@
...
+        requests = cstruct.get('requests', [])
+        for request in requests:
+            for k, v in defaults.items():
+                if k == 'headers':
+                    for i, j in v.items():
+                        request.get(k, {}).setdefault(i, j)
```
ametaireau: I like that :)

docs/batch.rst
Natim: Or ``request_defaults``

leplatrem: I'm not a fan to be honest :( ``defaults`` is fine IMO :)

```
``defaults``: (*optional*) default values for attributes of every batch request
```

docs/batch.rst
leplatrem: Will be adressed in #101

readinglist/views/batch.py
leplatrem: Refactored in #98 :)

docs/batch.rst
ametaireau: Reading the docs it's fine, but the name of the argument doesn't convey much more meaning, and I had to read the docs to understand what the purpose was.

Issue #84, opened 2015-02-04
Support default sort fields on resource


leplatrem:
Articles will be ordered by ``-stored_on`` by default (i.e. newest first).

Natim:
For pagination we want this to be ``-last_modified``and not configurable :)

PR #94, opened 2015-02-05
Implements Pagination (fixes #25)

### LINKED ISSUE #25: Pagination
#13 and #14 raise the question of how to fetch large numbers of items.

We can expect some number of users on day one to have hundreds of items. For illustration, there are a number of folks on the mobile team who have more than 1500 items in their current Pocket queue.

In ordinary operation — frequently connected — we can expect downloaded changesets to be small; a handful of new and changed items in each fetch.

In protracted disconnected operation we'll need batching for uploads, which is #2.

That leaves us worrying about first-time download on a device, and subsequent downloads after a very protracted disconnected period (or after an import).

We can approach this in one of a few ways.

1. Ignore the problem. Always maintain a full list of items on the client. Fetch everything in one go. After all, the metadata for even a thousand items will probably only be a few hundred KB. This is fine for initial development, but probably not enough for a global release — even a few hundred KB is too much in surprisingly many locales.

2. Ignore the problem except for downloads. Always maintain a full list of items on the client, but fetch items in batches. This is easy if the server supports sorting (already desirable: #14) and one more thing: range queries over some immutable attribute. E.g.,

```
articles/?sort=-stored_on&stored_on_less_than=123456
```

This solves some reliability/interrupted download issues, but still triggers significant space and bandwidth costs immediately after install. (What would you think if Gmail downloaded the headers for every message to your phone after signing in?)

3. Go fully server-canonical, and expect most clients to have only a partial cache of the articles that are stored on the server. For example, clients might not bother fetching metadata about archived articles until the user tries to see that list. Pocket falls into this category: it'll fetch new batches of articles as you scroll through your list.

This requires much the same support from the server, but imposes additional smarts on the client.

I think we should go with (1) during development, then shift to (2) or (3) as soon as we can.

Natim:


readinglist/backend/memory.py
```diff
@@ -78,12 +78,26 @@ def delete(self, resource, user_id, record_id):
         self._store[resource_name][user_id].pop(record_id)
         return existing
 
-    def get_all(self, resource, user_id, filters=None, sorting=None):
+    def get_all(self, resource, user_id, filters=None, sorting=None,
```
Natim: We could probably use a similar function for both simpleredis and memory.py

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+            limit = int(queryparams.get('_limit', settings.get(
+                'readinglist.paginate_by', DEFAULT_PAGINATE_BY)))
+        except ValueError:
+            error_details = {
+                'name': None,
+                'location': 'querystring',
+                'description': "_limit should be an integer"
+            }
+            self.request.errors.add(**error_details)
+            raise errors.json_error(self.request.errors)
```
leplatrem: We now have a ``self.raise_invalid`` :)

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+        if not last_modified_used:
+            sorting.append(('last_modified', -1))
         return sorting
 
+    def _extract_pagination_token(self, queryparams):
+        """Get pagination params."""
+        settings = self.request.registry.settings
+        try:
+            limit = int(queryparams.get('_limit', settings.get(
+                'readinglist.paginate_by', DEFAULT_PAGINATE_BY)))
```
leplatrem: split into two variables ? (better readability ihmo)

readinglist/resource.py
```diff
@@ -228,12 +234,17 @@ def _extract_filters(self, queryparams):
...
+                        (self.modified_field, value, COMPARISON.GT)
+                    )
+                elif param == '_to':
+                    filters.append(
+                        (self.modified_field, value, COMPARISON.LT)
+                    )
                 continue
 
-            m = re.match(r'^(min|max|not)_(\w+)$', param)
+            m = re.match(r'^(min|max|not|lt|gt)_(\w+)$', param)
```
leplatrem: This could be added to documentation then !

readinglist/resource.py
```diff
@@ -256,7 +267,8 @@ def _extract_sorting(self, queryparams):
         """Extracts filters from QueryString parameters."""
         specified = queryparams.get('_sort', '').split(',')
         sorting = []
-        for field in specified:
+        last_modified_used = False
+        for field in specified[:1]:
```
leplatrem: ``[:1]`` ?

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
                     }
                     self.raise_invalid(**error_details)
 
+                if field == 'last_modified':
```
leplatrem: self.modified_field

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+                    last_modified_used = True
+
                 direction = -1 if order == '-' else 1
                 sorting.append((field, direction))
 
+        if not last_modified_used:
+            sorting.append(('last_modified', -1))
```
leplatrem: self.modified_field

readinglist/resource.py
Natim: We only take the first sorting parameter.

readinglist/resource.py
```diff
@@ -256,7 +267,8 @@ def _extract_sorting(self, queryparams):
         """Extracts filters from QueryString parameters."""
         specified = queryparams.get('_sort', '').split(',')
         sorting = []
-        for field in specified:
+        last_modified_used = False
```
leplatrem: ``last_modified_used = self.modified_field in specified``

readinglist/resource.py
leplatrem: Add a comment, why -1 ?

Natim: By default it is ordered by ``-last_modified``

readinglist/resource.py
```diff
@@ -287,16 +398,28 @@ def collection_get(self):
 
         filters = self._extract_filters(self.request.GET)
         sorting = self._extract_sorting(self.request.GET)
-        records = self.db.get_all(filters=filters,
-                                  sorting=sorting,
-                                  **self.db_kwargs)
+        pagination_rules, limit = self._extract_pagination_token(
```
leplatrem: method name does not match returned values

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+                'location': 'querystring',
+                'description': "_limit should be an integer"
+            }
+            self.request.errors.add(**error_details)
+            raise errors.json_error(self.request.errors)
+
+        token = queryparams.get('_token', None)
+        filters = []
+        if token:
+            rules = json.loads(b64decode(token))
```
leplatrem: Could raise ValueError ?

readinglist/resource.py
```diff
@@ -287,16 +398,28 @@ def collection_get(self):
...
+        records, total_records = self.db.get_all(
+            filters=filters,
+            sorting=sorting,
+            pagination_rules=pagination_rules,
+            limit=limit,
+            **self.db_kwargs)
+
+        num_records = '%s' % total_records
+        self.request.response.headers['Total-Records'] = num_records.encode(
+            'utf-8')
```
leplatrem: What happened to you guys ? PEP8 is supposed to improve readability and flow !

```
num_records = ('%s' % total_records).encode('utf-8')
self.request.response.headers['Total-Records'] = num_records
```

readinglist/resource.py
```diff
@@ -287,16 +398,28 @@ def collection_get(self):
...
+
+        if len(records) == limit and total_records > limit:
+            self.request.response.headers['Next-Page'] = self._next_page_url(
+                sorting, limit, records[-1]).encode('utf-8')
```
leplatrem:
```
next_page = self._next_page_url(sorting, limit, records[-1])
self.request.response.headers['Next-Page'] = next_page.encode('utf-8')
```
?

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+
+        token = queryparams.get('_token', None)
+        filters = []
+        if token:
+            rules = json.loads(b64decode(token))
+            for rule in rules:
+                filters.append(self._extract_filters(rule))
+        return filters, limit
+
+    def _next_page_url(self, sorting, limit, last_record):
```
leplatrem: Missing small docstring

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+        queryparams = self.request.GET.copy()
+        queryparams['_limit'] = limit
+        queryparams['_token'] = self._build_pagination_token(
+            sorting, last_record)
+        return '%s%s?%s' % (self.request.host_url, self.request.path_info,
+                            urlencode(queryparams))
+
+    def _build_pagination_token(self, sorting, last_record):
+        """Build a continuation filtering rules.
```
leplatrem: *continuation token* ?

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+        try:
+            _, last_modified_direction = [x for x in sorting
+                                          if x[0] == 'last_modified'][0]
+        except IndexError:
+            last_modified_direction = 1
+
+        if last_modified_direction == -1:
+            last_modified_key = '_to'
+        else:
+            last_modified_key = '_since'
```
leplatrem: Since ``_to`` and ``_since`` appear several times in the code, I believe constants would make sense

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+
+            last_modified < 123
+
+        """
+        first_sort_field = None
+        last_sort_field = None
+        if len(sorting) > 1:
+            first_sort_field, first_sort_direction = sorting[0]
+            last_sort_field, last_sort_direction = sorting[-1]
+            if last_sort_field == 'last_modified':
```
leplatrem: self.modified_field

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+                last_sort_field, last_sort_direction = sorting[-2]
+
+        try:
+            _, last_modified_direction = [x for x in sorting
+                                          if x[0] == 'last_modified'][0]
```
leplatrem: ``itertools.ifilter(lambda x: x[0] == self.modified_field, sorting)[0]`` ?

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+                first_sort_key = 'gt_%s' % first_sort_field
+
+            filters1 = {first_sort_field: first_sort_value,
+                        last_modified_key: last_modified_value}
+
+            filters2 = {first_sort_key: first_sort_value}
+
+            for sort_field, sort_direction in \
+                    [x for x in sorting if x[0] not in
+                     ['last_modified', first_sort_field]]:
```
leplatrem: use an intermediary variable here

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+                value = '%s' % last_record.get(sort_field)
```
leplatrem: why do you convert the value to string ?

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+            first_sort_field, first_sort_direction = sorting[0]
+            last_sort_field, last_sort_direction = sorting[-1]
+            if last_sort_field == 'last_modified':
+                last_sort_field, last_sort_direction = sorting[-2]
+
+        try:
+            _, last_modified_direction = [x for x in sorting
+                                          if x[0] == 'last_modified'][0]
+        except IndexError:
+            last_modified_direction = 1
```
leplatrem: Instead of assigning ``-1`` and ``1`` and compare to ``1`` and ``-1`` in the code, **maybe** it would be readable to rename to ``last_modified_desc`` and assign with True or False.

readinglist/resource.py
Natim: Well you know we did some JS this year.

readinglist/resource.py
Natim: Well the we extract the pagination_token into pagination rules why is that wrong?

readinglist/resource.py
```diff
@@ -228,12 +234,17 @@ def _extract_filters(self, queryparams):
...
                     self.raise_invalid(**error_details)
 
-                filters.append(
-                    (self.modified_field, value, COMPARISON.GT)
-                )
+                if param == '_since':
+                    filters.append(
+                        (self.modified_field, value, COMPARISON.GT)
+                    )
+                elif param == '_to':
```
leplatrem: This could be added to documentation then !

readinglist/resource.py
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+        """Build a continuation filtering rules.
+
+        Something like that when sorting::
+
+            (status == 2 and last_modified < 123) OR status < 2
+
+
+        Something like that when sorting only per last_modified::
+
+            last_modified < 123
```
leplatrem: Maybe it would help if you write a small summary of what the code does. I try here, to check that I understood well :)

```
Build a continuation token, as a piece of JSON encoded in base64, that will be
used by the client to resume the current pagination state.

The goal of the algorithm is the build the set of filters (using min/max/gt/lt) that
will reduce the dataset of the next page. The combination of sorting and paginating
is complex to resolve for this problem. Depending on the direction of sorting for each
field, the set of filters will target values above or below last record field values.

#. Identify if we sort on last modified (use ``_since`` or ``_to`` relatively)
#. Loop on the sorting fields and their direction
#. Build a list of filters for each of them using ``gt`` or ``lt`` depending on the direction
#. Serialize the filters and encode to base64

````

File: readinglist/resource.py [hunk 23]
```diff
@@ -270,11 +282,110 @@ def _extract_sorting(self, queryparams):
...
+
+        """
+        first_sort_field = None
+        last_sort_field = None
```
leplatrem: why just first and last ?

(if it is a limitation, explicit it in docstring)

readinglist/backend/__init__.py
```diff
@@ -86,3 +88,23 @@ def apply_sorting(records, sorting):
         result = sorted(result, key=itemgetter(field), reverse=reverse)
 
     return result
+
+
+def handle_records_pagination(records, filters, sorting,
+                              pagination_rules=None, limit=None):
```
leplatrem: ``extract_record_set`` ?

readinglist/backend/__init__.py
```diff
@@ -86,3 +88,23 @@ def apply_sorting(records, sorting):
...
+    if not pagination_rules:
+        pagination_rules = []
+    filtered = list(apply_filters(records, filters or []))
+    total_records = len(filtered)
+    paginated = {}
+    for rule in pagination_rules:
+        values = list(apply_filters(filtered, rule))
+        paginated.update(dict(((x['_id'], x) for x in values)))
```
leplatrem: Give some line spacing, for a better reading flow

readinglist/backend/__init__.py
```diff
@@ -86,3 +88,23 @@ def apply_sorting(records, sorting):
...
+    if not paginated:
```
leplatrem: I would have inverted this test

readinglist/resource.py
Natim: I don't know which one is the most readable.

readinglist/resource.py
```diff
@@ -216,10 +220,12 @@ def _extract_filters(self, queryparams):
             param = param.strip()
             value = native_value(value)
 
-            if param == '_sort':
+            # Ignore specific fields
+            if param.startswith('_') and param not in ('_since', '_to'):
```
leplatrem: I don't understand this test

readinglist/resource.py
Natim: Because queryparams are all strings.

readinglist/resource.py
leplatrem: use intermediary variable then

readinglist/resource.py
leplatrem: but json.dumps isn't supposed to do this ?

readinglist/resource.py
leplatrem: Explicit why (limitation or whatever) with a comment

readinglist/resource.py
leplatrem: Add comment ;)

readinglist/resource.py
leplatrem: Ok, now that I understand the code better, it makes sense.

docs/api.rst
```diff
@@ -102,6 +105,10 @@ If the request header ``If-Modified-Since`` is provided, and if the
 collection has not suffered changes meanwhile, a ``304 Not Modified``
 response is returned.
 
+:note:
+   You can also use the ``_to`` parameter which works as an alias for
+   ``max_last_modified``
```
leplatrem: ``lt_last_modified`` !

readinglist/resource.py
```diff
@@ -273,8 +285,112 @@ def _extract_sorting(self, queryparams):
...
+            }
+            self.raise_invalid(**error_details)
+
+        token = queryparams.get('_token', None)
+        filters = []
+        if token:
+            try:
+                rules = json.loads(b64decode(token))
+            except (ValueError, TypeError):
+                pass
```
leplatrem: Don't we want to raise_invalid ?

readinglist/resource.py
```diff
@@ -287,16 +403,27 @@ def collection_get(self):
 
         filters = self._extract_filters(self.request.GET)
         sorting = self._extract_sorting(self.request.GET)
-        records = self.db.get_all(filters=filters,
-                                  sorting=sorting,
-                                  **self.db_kwargs)
+        pagination_rules, limit = self._extract_pagination_rules_from_token(
+            self.request.GET)
```
leplatrem: Since ``_extract_*`` methods have ``self`` maybe we could skip passing self.request.GET in arguments

readinglist/tests/resource/test_sort.py
```diff
@@ -48,38 +48,40 @@ def test_filter_errors_are_json_formatted(self):
...
+    #     self.resource.request.GET = {'_sort': 'status,-title'}
+    #     result = self.resource.collection_get()
+    #     self.assertEqual(result['items'][0]['status'], 0)
+    #     self.assertEqual(result['items'][0]['title'], 'MoFo #16')
+    #     self.assertEqual(result['items'][1]['status'], 0)
+    #     self.assertEqual(result['items'][1]['title'], 'MoFo #12')
+    #     self.assertEqual(result['items'][-2]['status'], 3)
+    #     self.assertEqual(result['items'][-2]['title'], 'MoFo #07')
+    #     self.assertEqual(result['items'][-1]['status'], 3)
+    #     self.assertEqual(result['items'][-1]['title'], 'MoFo #03')
```
leplatrem: why do you want to get rid of them ?

readinglist/resource.py
```diff
@@ -254,8 +265,9 @@ def _extract_filters(self, queryparams):
 
     def _extract_sorting(self, queryparams):
         """Extracts filters from QueryString parameters."""
-        specified = queryparams.get('_sort', '').split(',')
+        specified = queryparams.get('_sort', '').split(',')[:1]
```
leplatrem: You can't just put this and add no comment !

leplatrem: IHMO it is not fair to limit sorting to one field here.

You should take the first sorting parameter into account (and forget the other ones), only if paginating

readinglist/resource.py
```diff
@@ -273,8 +285,112 @@ def _extract_sorting(self, queryparams):
                 direction = -1 if order == '-' else 1
                 sorting.append((field, direction))
 
+        if not modified_field_used:
+            # Add a sort by the ``modified_field`` in descending order
+            # useful for pagination
+            sorting.append((self.modified_field, -1))
```
leplatrem: I don't agree with this here. In the docs we said it would be ``-stored_on`` (see #84)

If you need this, put it only if pagination is active

leplatrem: (i.e. in the pagination code, not here)

readinglist/tests/resource/test_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self.resource.known_fields = ['status', 'unread', 'title']
+        for i in indices:
+            record = {
+                'title': 'MoFo #{0:02}'.format(i),
+                'status': i % 4,
+                'unread': (i % 2 == 0)
+            }
+            self.db.create(self.resource, 'bob', record)
+        self.resource.request.GET = {}
```
leplatrem: redundant with basetest

readinglist/tests/resource/test_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+        self.last_response.headers = {}
```
leplatrem: I don't think this should be necessary

readinglist/tests/resource/test_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_return_items(self):
+        result = self.resource.collection_get()
+        self.assertEqual(len(result['items']), 20)
+
+    def test_handle_limit(self):
+        self.resource.request.GET = {'_limit': '10'}
+        result = self.resource.collection_get()
+        self.assertEqual(len(result['items']), 10)
+
+    def test_return_next_page_url(self):
```
leplatrem: ``test_next_page_url_is_given_in_headers``

readinglist/tests/resource/test_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertIn('Next-Page', self.last_response.headers)
+
+    def test_next_page_url_has_got_querystring(self):
+        self.resource.request.GET = {'_limit': '10'}
+        self.resource.collection_get()
+        queryparams = self._setup_next_page()
+        self.assertIn('_limit', queryparams)
+        self.assertIn('_token', queryparams)
+
+    def test_next_page_url_give_next_page(self):
```
leplatrem: gives

readinglist/tests/resource/test_pagination.py
```diff
@@ -0,0 +1,165 @@
...
+        self.resource.request.GET = {'_limit': '10'}
+        results1 = self.resource.collection_get()
+        self._setup_next_page()
+        results2 = self.resource.collection_get()
+        self.assertFalse(
+            set([x['_id'] for x in results1['items']]).intersection(
+                set([x['_id'] for x in results2['items']])))
```
leplatrem: using intermediary variables would help here

leplatrem:
Nice ! 

I suggest we add a few more tests that cover these parts of the code:

* pagination_rules in test_backend (small test)
* lt_ and gt_ in test_filter
* _to in test_sync
* _limit not an integer returns 400
* _token not base64
* _token not json
* Take out b64encode(json.dumps(token).encode('utf-8')).decode('utf-8') and add some unit tests for expected set of filters returned by _build_pagination_token. There are a lot of internal corner cases in this function, and it definitely deserves a dedicated test suite IHMO

readinglist/resource.py
Natim: Actually it doesn't for int values.

readinglist/resource.py
Natim: Yes

readinglist/tests/resource/test_sort.py
Natim: Because we don't handle multiple sorting anymore.

readinglist/resource.py
Natim: Well you are always paginating.

readinglist/resource.py
Natim: Well it just skips specific fields (that starts with '_') except `_since` and `_to` that are to be handled in this method. (It basically skips ``_sort``, ``_limit`` and ``_token``)

leplatrem: Ok I get it! Maybe in 6 months we will like to read it in the comment, no ? ``# Ignore specific fields like _sort _limit or _token``

readinglist/resource.py
```diff
@@ -273,8 +289,113 @@ def _extract_sorting(self, queryparams):
...
+        queryparams = self.request.GET
+        settings = self.request.registry.settings
+        paginate_by = settings.get('readinglist.paginate_by')
+        limit = queryparams.get('_limit', paginate_by)
+        try:
+            if limit:
+                limit = int(limit)
+        except ValueError:
+            error_details = {
+                'name': None,
```
leplatrem: name None is default and *could* be dropped here (maybe your prefer explicit)

leplatrem:
like @ametaireau  would say : oh yeah :) :+1:

readinglist/resource.py
Natim: Well this is after ``-stored_on`` which is not unique. We will handle default sortering in #84 this won't impact :)

PR #96, opened 2015-02-05
Increment timestamp on delete (fixes #95)

### LINKED ISSUE #95: Delete should increment last-modified on collection
http://pastebin.mozilla.org/8562034

leplatrem:


n1k0:
Tested against latest readinglist-client, solves the issue with Last-Modified header not being updated after a DELETE operation. Good job!

PR #69, opened 2015-02-09
FxA persist the state on server side instead of cookie.

Natim:


loadtests/server.ini
```diff
@@ -1,6 +1,7 @@
 [app:main]
 use = egg:readinglist
 
+readinglist.session = readinglist.sessions.redis
```
ametaireau: should we name this "sessionbackend" ?

ametaireau: "session_backend" even

readinglist/sessions/__init__.py
```diff
@@ -0,0 +1,24 @@
+class SessionStorageBase(object):
```
ametaireau: I believe there are a bunch of libs which are doing that. Among these ones, we have one from Ben, which is named baker, and is used for pyramid. https://pypi.python.org/pypi/pyramid_beaker.

Relevant?

config/readinglist.ini
```diff
@@ -10,9 +11,9 @@ readinglist.eos =
 readinglist.userid_hmac_secret = b4c96a8692291d88fe5a97dd91846eb4
 # readinglist.batch_max_requests = 25
 
-fxa-oauth.client_id =
-fxa-oauth.client_secret =
-fxa-oauth.oauth_uri = https://oauth.accounts.firefox.com/v1
+fxa-oauth.client_id = 89513028159972bc
+fxa-oauth.client_secret = 9aced230585cc0aa2932e2eb871c9a3a7d6458e59ccf57eb610ea0a3467dd800
+fxa-oauth.oauth_uri = https://oauth-stable.dev.lcip.org/v1
```
leplatrem: Should we ?

loadtests/server.ini
leplatrem: along with storage_backend then

config/readinglist.ini
Natim: Yes I think it will ease developer integration.

loadtests/server.ini
Natim: ok will do.

readinglist/sessions/__init__.py
leplatrem: +1 to use pyramid eco-system !
(too late ?)

readinglist/tests/test_session.py
```diff
@@ -0,0 +1,77 @@
...
+        retrieved = self.session.get('foobar')
+        self.assertEquals(retrieved, stored)
+
+    def test_expire_expires_the_value(self):
+        stored = 'toto'
+        self.session.set('foobar', stored)
+        self.session.expire('foobar', 0.05)
+        time.sleep(0.1)
+        retrieved = self.session.get('foobar')
+        self.assertIsNone(retrieved)
```
leplatrem: nice!

readinglist/tests/test_views_oauth.py
```diff
@@ -37,32 +48,32 @@ def test_fails_if_state_or_code_is_missing(self):
...
 
     def test_fails_if_state_does_not_match(self):
+        self.app.app.registry.session.set('def', 'http://foobar')
         url = '{url}?state=abc&code=1234'.format(url=self.url)
-        headers = {'Cookie': 'state=def'}
-        self.app.get(url, headers=headers, status=400)
+        self.app.get(url, status=401)
 
     @mock.patch('readinglist.views.oauth.OAuthClient.trade_code')
     def tests_returns_token_traded_against_code(self, mocked_trade):
```
leplatrem: the test should be renamed to say we expect a redirection

readinglist/views/oauth.py
```diff
@@ -74,12 +75,12 @@ def fxa_oauth_token(request):
...
-    return data
+    except fxa_errors.InProtocolError as error:
+        logger.exception(error)
+        request.errors.add(
+            location='querystring',
+            name='code',
+            description='Firefox Account code validation failed.')
+        raise json_error(request.errors)
+
+    return HTTPFound(location='%s?token=%s' % (stored_redirect, token))
```
leplatrem: That change might required some documentation update.
Are we sure that we want that ? Will it be easier for clients ?

readinglist/sessions/__init__.py
Natim: For know we don't use session with a ``session_id`` it seems a bit to much for what we are planning to do. But if we want to add session and store the session_id in a cookie, we will definitely consider using this.

readinglist/views/oauth.py
Natim: For know they actually need this it won't work for them as it is now.

Natim: Documentation **updated**.

docs/authentication.rst
```diff
@@ -19,16 +19,7 @@ Use the OAuth token with this header:
...
-  access ``GET /v1/fxa-oauth/token``, where its session will be
-  validated. Then, an OAuth token will be returned inside a JSON
-  object:
-
-::
-
-    {
-        "token": "oihyvuh-fefe-ldpieo98963fyhrn"
-    }
+  be redirected to ``http://app-endpoint?token={token}`` the web-app.
```
leplatrem: *on* the web-app?

Does web-app cover mobile native clients ?

What would be app-endpoint with mobile native apps ?

readinglist/views/oauth.py
leplatrem: Wouldn't it be easier for http clients if we add a header ? 
Parsing the location header can be painful and error-prone no ?

docs/authentication.rst
Natim: On Firefox OS you can define a catch on redirection to ``app://readinglist/fxa-oauth/token`` on Android I don't know but it works the same for all Oauth flow.

readinglist/views/oauth.py
Natim: I don't think web-apps can handle the header at this point of the flow.

PR #103, opened 2015-02-09
Consistency for unique fields definition

leplatrem:
Idea... unicity constraints shall also apply to id and last-modified (*backend constraints consistency*)


Natim:
Then we could put back the PUT endpoint :)

PR #106, opened 2015-02-09
Handle deleted items

### LINKED ISSUE #41: Include deleted items in articles collection
* Deleted items should appear in ``/articles`` when polling for changes (ie. ``?_since=1234``)
* Deleted items appear with status=2
* Deleted items have empty values for read position, marked_(by|on), favorite
* URL of deleted items should not interfere with URL unicity (ref #20)
* ~~Article status cannot be modified to value 2~~ done in #77

leplatrem:
Fixes #41 

* [x] Basic behaviour, strategy and unit tests
* [x] Redis backend implem
* [x] Test that deleted articles have no attributes anymore (already test in backend)
* [x] Test that filtering works on status=2
* [x] Test that sorting works on deleted in resource tests
* [x] Test that sorting works on status in article tests
* [x] Test that unicity for deleted items does not interfere
* [x] Update docs
* [x] Update load tests
* [ ] ~~Find a strategy to purge deleted items~~ #107

### DIFF
--- a/readinglist/storage/memory.py
+++ b/readinglist/storage/memory.py
@@ -72,10 +72,10 @@ class Memory(StorageBase):
     def delete(self, resource, user_id, record_id):
         resource_name = classname(resource)
         existing = self.get(resource, user_id, record_id)
-        self.set_record_deleted_mark(resource, user_id, existing)
-        self.set_record_timestamp(resource, user_id, existing)
+        record = self.strip_deleted_record(resource, user_id, existing)
+        self.set_record_timestamp(resource, user_id, record)
         self._store[resource_name][user_id].pop(record_id)
-        return existing
+        return record
 
     def get_all(self, resource, user_id, filters=None, sorting=None,
                 pagination_rules=None, limit=None):
--- a/readinglist/storage/memory.py
+++ b/readinglist/storage/memory.py
@@ -16,7 +16,7 @@ class Memory(StorageBase):
 
     def flush(self):
         self._store = tree()
-        self._cementery = tree()
+        self._cemetery = tree()
         self._timestamps = defaultdict(dict)
 
     def ping(self):
@@ -76,7 +76,7 @@ class Memory(StorageBase):
         self.set_record_timestamp(resource, user_id, existing)
 
         # Add to deleted items, remove from store.
-        self._cementery[resource_name][user_id][record_id] = existing.copy()
+        self._cemetery[resource_name][user_id][record_id] = existing.copy()
         self._store[resource_name][user_id].pop(record_id)
 
         self.strip_deleted_record(resource, user_id, existing)
@@ -89,7 +89,7 @@ class Memory(StorageBase):
 
         deleted = {}
         if include_deleted:
-            deleted = self._cementery[resource_name][user_id]
+            deleted = self._cemetery[resource_name][user_id]
 
         records, count = extract_record_set(records + deleted.values(),
                                             filters, sorting,
--- a/readinglist/storage/memory.py
+++ b/readinglist/storage/memory.py
@@ -85,14 +85,14 @@ class Memory(StorageBase):
     def get_all(self, resource, user_id, filters=None, sorting=None,
                 pagination_rules=None, limit=None, include_deleted=False):
         resource_name = classname(resource)
-        records = self._store[resource_name][user_id].values()
+        records = list(self._store[resource_name][user_id].values())
 
         deleted = {}
         if include_deleted:
             deleted = self._cemetery[resource_name][user_id]
 
         records, count = extract_record_set(resource,
-                                            records + deleted.values(),
+                                            records + list(deleted.values()),
                                             filters, sorting,
                                             pagination_rules, limit)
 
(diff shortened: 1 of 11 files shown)


leplatrem:
r? @Natim @ametaireau 

(Redis backend is not implemented yet, waiting for your feedback on strategy before going further)

readinglist/backend/__init__.py
```diff
@@ -36,6 +36,16 @@ def get_all(self, resource, user_id, filters=None, sorting=None,
                 pagination_rules=None, limit=None):
         raise NotImplementedError
 
+    def strip_deleted_record(self, resource, user_id, record):
+        """Strip the record to its deleted version (in place)."""
```
ametaireau: I don't get what this means from the comment.

It seems to remove all the fields of the record which aren't the id or the modification field. Is that correct?

leplatrem: Yes, it strips (in place) all the fields that aren't id or last-modified. And then it has the deleted mark defined at the resource level

ametaireau: Okay, I think I got it: so when we delete something, we add a deletion marker on it and remove all the information associated with the record in the database.

readinglist/backend/memory.py
```diff
@@ -75,16 +76,33 @@ def update(self, resource, user_id, record_id, record):
...
-                pagination_rules=None, limit=None):
+                pagination_rules=None, limit=None, include_deleted=False):
         resource_name = classname(resource)
         records = self._store[resource_name][user_id].values()
-        return extract_record_set(records, filters, sorting,
-                                  pagination_rules, limit)
+
+        deleted = {}
+        if include_deleted:
+            deleted = self._cementery[resource_name][user_id]
```
ametaireau: Vouliez-vous dire cemetery?

readinglist/backend/__init__.py
Natim: I would rather use a deleted_items hash table per user with ``{"id": <deletion_last_modified>}`` to avoid having deleted items mixed with used ones.

Natim: What you seems to do with ``cementery``

readinglist/views/article.py
```diff
@@ -67,6 +67,7 @@ class Options:
 @crud()
 class Article(BaseResource):
     mapping = ArticleSchema()
+    deleted_mark = ('status', 2)
```
Natim: Actually I'd like to do both ``deleted: True`` and ``status: 2`` and if possible ask @rnewman to get rid of this ``status: 2`` thing since we cannot patch on it and it is less explicit than having ``deleted: True``

leplatrem: What's wrong with ``status : 2`` ?

We couldn't patch on ``deleted`` either!

You remind me of something : we may want to filter on status=2 :) I'll add a test to check that it works :)

readinglist/backend/__init__.py
leplatrem: > to avoid having deleted items mixed with used ones

If I'm not mistaken, we want that...

readinglist/views/article.py
Natim: It is exactly why I don't want deleted items to be mixed with items to prevent user from trying to sort or paginate on them.

leplatrem: Why wouldn't you want to obtain only the deleted items since timestamp X ?

Natim: Did I said that?

leplatrem: > It is exactly why I don't want deleted items [...] to prevent user from trying to sort or paginate on them.
Ok, sort and paginate, not filter you're right.

I shall add an assertion then, preventing to sort on resource.deleted_mark

readinglist/storage/memory.py
```diff
@@ -72,16 +73,33 @@ def update(self, resource, user_id, record_id, record):
...
+
+        records, count = extract_record_set(resource,
+                                            records + list(deleted.values()),
+                                            filters, sorting,
+                                            pagination_rules, limit)
+
+        filtered_deleted = len([r for r in records
+                                if r[resource.id_field] in deleted])
+
+        return records, count - filtered_deleted
```
ametaireau: So, in case we return the deleted items, we don't count them?

Natim: There are not part of the collection since they've been deleted.

ametaireau: Well, we're returning them and paginating trough them, correct?

Natim: We are not using the total_records information for that pagination.

ametaireau: yeah, but I find it weird to have a number of items returned different than the actual number of items.

Natim: We are not returning the number of items in the list but just the number of items in the collection. Why would you need the number of items in the list? Isn't ``len(items)`` enough for that?

leplatrem: Indeed, distinguishing number of items in the results set from the size of the items list was necessary for pagination.

However, I decided that deleted items should not be counted in the results set, but it was a spontaneous decision... Appart from what you said @Natim (*they've been deleted*), I cannot see any strong argument !

PR #109, opened 2015-02-10
POST on /articles doesn't check for authentication.

Natim:
http POST http://localhost:8000/v0/articles \
        added_on=1423557413184 added_by=Natim \
        title=Ionyse url=http://www.ionyse.fr/

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -51,8 +51,8 @@ def set_auth(config):
     authn_policy = MultiAuthenticationPolicy(policies)
     authz_policy = authentication.AuthorizationPolicy()
 
-    config.set_authentication_policy(authn_policy)
     config.set_authorization_policy(authz_policy)
+    config.set_authentication_policy(authn_policy)
 
 
 def attach_http_objects(config):
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -51,8 +51,8 @@ def set_auth(config):
     authn_policy = MultiAuthenticationPolicy(policies)
     authz_policy = authentication.AuthorizationPolicy()
 
-    config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
+    config.set_authorization_policy(authz_policy)
 
 
 def attach_http_objects(config):
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -8,6 +8,7 @@ from pyramid.config import Configurator
 from pyramid.events import NewRequest, NewResponse
 from pyramid.httpexceptions import HTTPTemporaryRedirect
 from pyramid_multiauth import MultiAuthenticationPolicy
+from pyramid.security import NO_PERMISSION_REQUIRED
 
 from readinglist import authentication
 from readinglist.utils import msec_time
@@ -36,7 +37,8 @@ def handle_api_redirection(config):
                      pattern='/{path:(?!%s).*}' % API_VERSION)
 
     config.add_view(view=_redirect_to_version_view,
-                    route_name='redirect_to_version')
+                    route_name='redirect_to_version',
+                    permission=NO_PERMISSION_REQUIRED)
 
     config.route_prefix = '/%s' % API_VERSION
 
@@ -53,6 +55,7 @@ def set_auth(config):
 
     config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
+    config.set_default_permission('readwrite')
 
 
 def attach_http_objects(config):
--- a/readinglist/authentication.py
+++ b/readinglist/authentication.py
@@ -29,7 +29,6 @@ def check_credentials(username, password, request):
     userid = hmac.new(hmac_secret,
                       credentials.encode('utf-8'),
                       hashlib.sha256).hexdigest()
-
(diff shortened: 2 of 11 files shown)


Natim:
@leplatrem r?

ametaireau:
So, I'm leaving a comment here just for reference later. 

The problem lies in the fact that cornice resources are attaching information to the functions themselves, especially about the permissions. This information is not copied over in case of heritage, and the permissions are lost.

The best way to fix this is to change how cornice stores this information, and probably use a cleaner design than what's done currently, potentially by storing the permissions inside the object directly rather than attached to the functions.

readinglist/views/article.py
```diff
@@ -118,6 +119,7 @@ def preprocess_record(self, new, old=None):
 
         return new
 
+    @resource.view(permission='readwrite')
```
leplatrem: ouch

readinglist/tests/test_views_article.py
```diff
@@ -15,6 +15,18 @@ def setUp(self):
...
+                   'Content-Type': 'application/json'}
+        self.app.get('/articles', status=401, headers=headers)
+        self.app.post_json('/articles', MINIMALIST_ARTICLE, status=401,
+                           headers=headers)
+
+        self.app.get(self.url, status=401, headers=headers)
+        self.app.patch_json(self.url, MINIMALIST_ARTICLE, status=401,
+                            headers=headers)
+        self.app.delete(self.url, status=401, headers=headers)
+
```
leplatrem: I suggest we move this test outside modification (``IntegrationTest(BaseWebTest, TestCase)`` ?)

readinglist/__init__.py
```diff
@@ -51,8 +51,8 @@ def set_auth(config):
     authn_policy = MultiAuthenticationPolicy(policies)
     authz_policy = authentication.AuthorizationPolicy()
 
-    config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
+    config.set_authorization_policy(authz_policy)
```
leplatrem: I once switched those two... while trying to initialize an app from scratch in tests.resource.test_views

leplatrem:
I also suggest we set a default permission and disable explicitly public endpoints
 http://docs.pylonsproject.org/docs/pyramid/en/latest/narr/security.html#setting-a-default-permission

readinglist/tests/resource/test_views.py
```diff
@@ -57,11 +57,11 @@ class AuthzAuthnTest(BaseWebTest):
     def test_all_views_require_authentication(self):
         self.app.get(self.collection_url, status=401)
 
-        self.app.post(self.collection_url, MINIMALIST_RECORD, status=401)
+        self.app.post_json(self.collection_url, MINIMALIST_RECORD, status=401)
 
         url = self.get_item_url('abc')
         self.app.get(url, status=401)
-        self.app.patch(url, MINIMALIST_RECORD, status=401)
+        self.app.patch_json(url, MINIMALIST_RECORD, status=401)
```
leplatrem: :+1: good catch

readinglist/views/article.py
leplatrem: you should add it to put() as well. It should not return 405 if the user is not authenticated

Natim: Well I am not convinced about that last statement.

leplatrem:
> The problem lies in the fact that cornice resources are attaching information to the functions 
> themselves, especially about the permissions. This information is not copied over in case of
> heritage, and the permissions are lost.

Can be seen as a feature as well, to override default permissions defined in ``resource.py``

readinglist/views/article.py
Natim: You shouldn't need to be connected to know the behavior does not exists.

leplatrem: Well if not connected it will trigger 404 anyway I believe

readinglist/__init__.py
```diff
@@ -53,6 +55,7 @@ def set_auth(config):
 
     config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
+    config.set_default_permission('readwrite')
```
leplatrem: Should it be one of the specific permissions used in resource ? Or should it an abstract one like "view", that we override in resource views to readonly/readwrite ?

readinglist/authentication.py
```diff
@@ -79,7 +78,7 @@ def _get_credentials(self, request):
             raise HTTPServiceUnavailable()
         except (fxa_errors.InProtocolError, fxa_errors.TrustError):
             return None
-        return user_id
+        return 'fxa_%s' % user_id
```
leplatrem: Even if it is not a big deal, this will deserve a note in the changelog : all records lost ;)

readinglist/__init__.py
Natim: I give by default the most needed rights so that developers can ask for less permissions if needed.

readinglist/views/article.py
Natim: Tested in 2c77537

readinglist/authentication.py
Natim: Added the CHANGELOG modification.

PR #117, opened 2015-02-11
Add deprecation indicators.

ametaireau:


docs/versionning.rst
```diff
@@ -21,19 +21,17 @@ Deprecation
...
+* ``code``: one of the strings ``"deprecated-client"``, ``"soft-eol"`` or ``"hard-eol"``;
+* ``message``: a human-readable message;
+* ``url``: a URL at which more information is available.
 
-.. A ``513 Service Decommissioned`` error response can be returned
-.. indicating that the service has been replaced with a new and better
-.. service using some as-yet-undesigned protocol.
+A ``410 Gone`` error response can be returned if the
+client version is too old, or the service had been remplaced with
+a new and better service using some as-yet-undesigned protocol.
```
Natim: ``some as-yet-undesigned protocol.``

a new and better service using the new protocol version.

readinglist/__init__.py
```diff
@@ -78,7 +84,12 @@ def on_new_request(event):
     config.add_subscriber(on_new_request, NewRequest)
 
     def on_new_response(event):
-        # Display the status of the request as well as the time spend
+        if hasattr(event.request, '_received_at'):
+            duration = (msec_time() - event.request._received_at)
+        else:
+            duration = "unknown"
```
Natim: This should not happen. Every request has to be receive at some point.

ametaireau: Well, it does! When you create a response directly then the event is triggered and the "event.request" associated hasn't passed the on_new_request (because, it's not a new request).

docs/versionning.rst
```diff
@@ -21,19 +21,17 @@ Deprecation
...
 
-.. Using the ``Alert`` header, the server can communicate any potential warning
-.. messages, information, or other alerts.
-.. The value is JSON mapping with the following attributes:
+The date of the end of support is provided in the API root URL (e.g. ``/v0``)
 
-.. * ``code``: one of the strings ``"deprecated-client"``, ``"soft-eol"`` or ``"hard-eol"``
-.. * ``message``: a human-readable message
-.. * ``url``: a URL at which more information is available
+Using the ``Alert`` header, the server can communicate any potential warning
```
leplatrem: response header

docs/versionning.rst
```diff
@@ -21,19 +21,17 @@ Deprecation
...
+messages, information, or other alerts.
+The value is JSON mapping with the following attributes:
```
leplatrem: a JSON mapping ?

docs/versionning.rst
leplatrem: using a new protocol version

readinglist/errors.py
```diff
@@ -23,7 +24,8 @@
     REQUEST_TOO_LARGE=113,
     CLIENT_REACHED_CAPACITY=117,
     UNDEFINED=999,
-    BACKEND=201
+    BACKEND=201,
+    END_OF_LIFE=401
```
leplatrem: Is it listed in docs/errors.rst?

readinglist/errors.py
```diff
@@ -85,6 +87,23 @@ def __init__(self, **kwargs):
         super(HTTPInternalServerError, self).__init__(**kwargs)
 
 
+class HTTPServiceDeprecated(PyramidHTTPGone):
+    """A HTTPGone formatted in JSON."""
+
+    def __init__(self, **kwargs):
+        kwargs.setdefault('body', format_error(
+            410,
```
leplatrem: In other parts of the project, we used PyramidHTTPGone.code and PyramidHTTPGone.title

docs/versionning.rst
ametaireau: What's wrong with a JSON mapping?

readinglist/__init__.py
```diff
@@ -97,9 +108,36 @@ def on_new_response(event):
...
 
 
+def end_of_life_tween_factory(handler, registry):
+    """Pyramid tween to handle service end of life."""
+
+    def eos_tween(request):
+        eos_date = registry.settings.get("readinglist.eos")
+        eos_url = registry.settings.get("readinglist.eos_url")
+        if eos_date:
+            eos_date = dateparser.parse(eos_date)
```
leplatrem: Could this go out of the tween (using the closure) to prevent parsing setting at each request ?

readinglist/__init__.py
```diff
@@ -97,9 +108,36 @@ def on_new_response(event):
...
+                alert['url'] = eos_url
+
+            if eos_date > datetime.datetime.now():
+                alert['code'] = "soft-eol"
+                response = handler(request)
+            else:
+                response = HTTPServiceDeprecated()
+                alert['code'] = "hard-eol"
+                response.headers = {}
+            response.headers['Alert'] = json.dumps(alert)
```
leplatrem: You said in the docs there will be a ``message``, maybe it should be mentionned as optional

readinglist/errors.py
ametaireau: I can reuse the code yeah. The title will not stay the same, since I changed it to "Service deprecated".

readinglist/__init__.py
ametaireau: Then I would need to find a better way to test it: currently the app is created once and the configuration patched afterwards (once the tween had been registered already). Thoughts?

leplatrem: Ah, I see :(

Alternatives are:
* create a new app with some settings that have eol enabled
* unit test of this factory and simply test that tween is registered
* use a "cache" in the closure to do it once only
* keep it this way :)

ametaireau: * Creating a new app with different settings would mean creating two different apps because I'm testing with a date in the past and in the future (and with no date)
* Unit testing wouldn't be enough here since I found some problems only when having it enabled within the scope of the app.
* I don't get what the cache proposal you mention is (it seems that if we cache in the closure we have the same problem, but I may miss something)

readinglist/__init__.py
ametaireau: I fixed this, and I'm now adding the message if it's present.

readinglist/__init__.py
leplatrem: The cache proposal would be like the closure expect that it would be set on the first request hit only. If you mock before the first request is hit, then you win :) Don't worry if you like the unit test way :)

PR #118, opened 2015-02-11
Enable Mocha reporter in make tests-once

leplatrem:
Sorry... couldn't resist :)


ametaireau:
Let's face it, we were using it already!

PR #119, opened 2015-02-12
Default permissions override OPTIONS permissions.

Natim:
Cornice need a patch, see https://github.com/mozilla-services/cornice/issues/273

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -55,7 +55,6 @@ def set_auth(config):
 
     config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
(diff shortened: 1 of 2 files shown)


readinglist/__init__.py
```diff
@@ -55,7 +55,6 @@ def set_auth(config):
 
     config.set_authorization_policy(authz_policy)
     config.set_authentication_policy(authn_policy)
-    config.set_default_permission('readwrite')
```
leplatrem: Instead of removing the line, would you mind leaving it with a XXX comment that it should be restored once the patch is merged and released upstream ?

Natim: Done

PR #124, opened 2015-02-16
Fixed fxa-oauth.oauth_uri.

n1k0:
The default configuration value for `fxa-oauth.oauth_uri` has an erroneous `/v1` path set; this patch fixes that.


ametaireau:
Thanks!

Issue #121, opened 2015-02-13
Authenticated requests are slow


n1k0:
![screen shot 2015-02-13 at 08 02 07](https://cloud.githubusercontent.com/assets/41547/6183580/f60c42b6-b356-11e4-9950-abbbf828ced1.png)

Natim:
Yes we probably need to cache the Bearer Token verification. cc @rfk @ckarlof

ckarlof:
We are working on our longer term story around this. It will likely involve some caching policy on the your side along with a notification system to inform you and other service providers of revocations and other events (https://github.com/mozilla/fxa-notification-server/blob/master/docs/api.md). This is a Q1 goal for us. 

Some additional information around this would be useful:
* I assume this is against our dev integration environment, or is this running locally?
* Can you please determine what the authentication overhead is, as measured by the RL server? It would be nice to compare this FxA dev vs FxA prod (or stage), too.

You could cache the tokens after you verify them on your side, but we need to figure out what the revocation story is. Caching them for some limited period of time will probably be a good short term tradeoff. We could also add a caching layer on the OAuth server for recently verified tokens. We should get some profiling data on the auth overhead first, though. Thoughts @rfk, @seanmonstar?

seanmonstar:
I agree, it'd be nice to see the timing on the server side. Is that request doing more than verification?

ckarlof:
@seanmonstar, in the meantime, we can also instrument `/verify` and see what the profile looks like there.

rfk:
Yes, in the short-term caching will be the right thing here.  @Natim it would be great if you can confirm that most of the overhead here is in talking to the FxA oauth verifier.

If so, let's add a function/class/whatever for doing oauth token verifications in PyFxA, and have it take care of the caching internally.  Something like a 5min cache timeout to start is probably sufficient, and we can iterate on this class to have it e.g. consume revocation notifications in the future without having to change client code.

Natim:
@rfk As you can see with PR #125 we are going from 1600ms per call to 2ms per call when enabling 5 minutes caching.

rfk:
> we are going from 1600ms per call to 2ms per call when enabling 5 minutes caching 

Heh, great to hear and no great surprise.  We need to firm up the story here on our side.

Natim:
Handled in https://github.com/mozilla-services/cliquet/pull/10

PR #125, opened 2015-02-17
Add a cache for FxA tokens validation.

Natim:
Refs #121


readinglist/authentication.py
```diff
@@ -64,20 +64,25 @@ def _get_credentials(self, request):
...
+
+            auth_client = OAuthClient(server_url=server_url)
+            try:
+                profile = auth_client.verify_token(token=auth, scope=scope)
+                hmac_secret = settings.get('readinglist.userid_hmac_secret')
+                user_id = hmac.new(hmac_secret.encode('utf-8'),
+                                   profile['user'].encode('utf-8'),
+                                   hashlib.sha256).hexdigest()
+                request.registry.session.set('fxa-oauth-token.%s' % auth, user_id)
+                request.registry.session.expire('fxa-oauth-token.%s' % auth, 5 * 60)
```
Natim: Needs to be in a settings.

n1k0:
Tested with latest client, approved.

config/readinglist.ini
```diff
@@ -15,6 +15,7 @@ fxa-oauth.client_id = 89513028159972bc
 fxa-oauth.client_secret = 9aced230585cc0aa2932e2eb871c9a3a7d6458e59ccf57eb610ea0a3467dd800
 fxa-oauth.oauth_uri = https://oauth-stable.dev.lcip.org
 fxa-oauth.scope = profile
+fxa-oauth.cache_verification = 300
```
ametaireau: cache_ttl

readinglist/authentication.py
```diff
@@ -64,20 +64,33 @@ def _get_credentials(self, request):
...
-            profile = auth_client.verify_token(token=auth, scope=scope)
-            hmac_secret = settings.get('readinglist.userid_hmac_secret')
-            user_id = hmac.new(hmac_secret.encode('utf-8'),
-                               profile['user'].encode('utf-8'),
-                               hashlib.sha256).hexdigest()
-        except fxa_errors.OutOfProtocolError:
-            raise HTTPServiceUnavailable()
-        except (fxa_errors.InProtocolError, fxa_errors.TrustError):
-            return None
+        session = getattr(request.registry, 'session', None)
```
ametaireau: can we pass the session directly when we're creating the authentication object?

config/readinglist.ini
n1k0: Nit: it's often considered good practice to name the setting using the type of unit used; here, `cache_ttl_seconds` would be meaningful.

leplatrem:
Small test to assert that fxa server is not hit during ttl ?

PR #129, opened 2015-02-18
Expose cors headers so clients subject to it can access it.

n1k0:
I wanted to add a test, but that's an exercise I should leave to you guys ;)


n1k0:
Updated to add support for `Total-Records` as well. This is obviously not the right way to handle this, though I can't seem being capable of emitting `Access-Control-Request-Headers` for preflight requests client-side, so in the meanwhile, I'd need this…

ametaireau:
We need to do this depending on the service, rather than doing it for all the services. Also, this should go in cliquet resources rather than in the readinglist now.

n1k0:
@ametaireau sure ; I could spend some time learning about that, but I suspect I'd rather need to focus on client dev instead. Thoughts?

Natim:
@n1k0 Don't worry on the server side we will take care of it as soon as we have clear the path with the on-going split (between readinglist and cliquet)

PR #133, opened 2015-02-19
Cliquet integration

ametaireau:


readinglist/__init__.py
```diff
@@ -29,135 +13,12 @@
...
-
-    set_auth(config)
 
     # Include cornice and discover views.
-    config.include("cornice")
-    config.scan("readinglist.views")
-
-    attach_http_objects(config)
+    config.include("cliquet", route_prefix=API_VERSION)
+    config.route_prefix = API_VERSION
```
leplatrem: Could not be done in include step ?

Natim: Unfortunatelly this is the only thing that is reset through include(). :(

readinglist/views/article.py
```diff
@@ -3,10 +3,10 @@
...
 from pyramid import httpexceptions
 
-from readinglist import errors
-from readinglist.resource import crud, BaseResource, ResourceSchema, TimeStamp
-from readinglist.utils import strip_whitespace
-from readinglist.schema import URL
+from cliquet import errors
+from cliquet.resource import crud, BaseResource, ResourceSchema, TimeStamp
+from cliquet.utils import strip_whitespace
+from cliquet.schema import URL
```
leplatrem: URL schema could have stayed in RL :)

readinglist/__init__.py
leplatrem: Ok

leplatrem:
LGTM :) Not many green lines in diff :)

readinglist/views/article.py
Natim: We've discussed that and it seems to make sense to have it in cliquet in case we want to reuse it in other projects don't it?

PR #139, opened 2015-02-26
Make sure to update dependencies on make serve.

Natim:
r? @leplatrem


leplatrem:
Are we sure to want that ? At each make serve  ?

Natim:
I think it is better than to ``make serve`` an outdated version. Also we can add ``make update`` to do so.

leplatrem:
No, I guess it's ok. It will make sense once we have proper cliquet versions. I merge!

PR #142, opened 2015-02-26
Backport Cliquet docs

leplatrem:
Supposed to ease future backporting :)

### DIFF
--- a/docs/resource.rst
+++ b/docs/resource.rst
@@ -1,21 +1,19 @@
-#################
-API Documentation
-#################
+##################
+Resource endpoints
+##################
 
-.. _http-apis:
+.. _resource-endpoints:
 
 GET /articles
 =============
 
-**Requires an FxA OAuth authentication**
+**Requires authentication**
 
-Returns all articles of the current user.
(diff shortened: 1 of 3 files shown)


Natim:
Are you sure do you want to do this modification in readinglist ?

docs/resource.rst
```diff
@@ -1,21 +1,19 @@
...
+.. _resource-endpoints:
 
 GET /articles
 =============
 
-**Requires an FxA OAuth authentication**
+**Requires authentication**
 
-Returns all articles of the current user.
+Returns all records of the current user for this resource.
```
Natim: Can we keep articles?

leplatrem:
My idea (suggestion) here was to backport as much docs as possible from Cliquet.
This way, when we upgrade Cliquet, the diff on docs is much smaller...

docs/resource.rst
leplatrem: I don't see why mentionning records would be a problem. If we keep articles, then this PR is not so relevant :)

PR #159, opened 2015-03-10
Centos 7 build changes

Micheletto:
You probably don't want to merge this as is, but I wasn't able to build on Centos 7 without upgrading pip to a version that supports the (now deprecated) --process-dependency-links option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,9 @@ $(INSTALL_STAMP): $(PYTHON) setup.py
 	$(VENV)/bin/pip install --process-dependency-links -U -e .
 	touch $(INSTALL_STAMP)
 
-install-dev: install $(DEV_STAMP)
-$(DEV_STAMP): $(PYTHON) dev-requirements.txt
-	$(VENV)/bin/pip install -r dev-requirements.txt
-	touch $(DEV_STAMP)
+install-dev:
+        $(VENV)/bin/pip install -r dev-requirements.txt
+        touch $(DEV_STAMP)
 
 virtualenv: $(PYTHON)
 $(PYTHON):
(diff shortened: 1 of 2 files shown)


dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
+pip==6.0.8
```
tarekziade: Maybe we can use >= here so it's future proof (since these dependencies are just for dev)

Makefile
```diff
@@ -19,10 +19,9 @@ $(INSTALL_STAMP): $(PYTHON) setup.py
 	$(VENV)/bin/pip install --process-dependency-links -U -e .
 	touch $(INSTALL_STAMP)
 
-install-dev: install $(DEV_STAMP)
-$(DEV_STAMP): $(PYTHON) dev-requirements.txt
-	$(VENV)/bin/pip install -r dev-requirements.txt
-	touch $(DEV_STAMP)
+install-dev: virtualenv
+        $(VENV)/bin/pip install -r dev-requirements.txt
```
Natim: Please don't remove the ``$(DEV_STAMP)`` rule, it is our way to not replay the install if not needed.

Natim:
Actually we don't have dependency_links anymore so we can just remove the argument I guess.

ametaireau:
Yes, we should just remove the argument.

Also, it seems weird to me to use the `install-dev` requirement for prod. What you want is just to run the `python setup.py install`, and that should work just fine.

leplatrem:
Maybe we should maintain the dependency_links feature, even if we don't use *right now*. I'm thinking of hot fixes in upstream libraries :)

ametaireau:
Closing this issue for now, we're tracking the installation on CentOS with bobm otherwise.

PR #16, opened 2015-03-12
Service provider with FxA

Natim:
Scope ``readinglist:`` ?


rfk:
@natim if you need the email address, yes, you'll need at least "profile:email" scope as well as "readinglist".  But consider whether you actually need it, because that one *is* PII that needs careful handling for storage etc.

Natim:
Yes I double checked and we are not asking for the email or the avatar. I think it will be fine with the readinglist only.

@rnewman I plan to change the configuration on the dev server can I ?

rnewman:
Fire away!

Natim:
Ok I am hitting 401 on the readinglist server when using the readinglist bearer token.

Natim:
Ok it works :)

PR #168, opened 2015-03-13
Updating documentation link in tests

pdehaan:
Per change in https://github.com/mozilla-services/readinglist/commit/faab111a9d4b204d2fcb5acc8cb1da327cc1d83a


leplatrem:
I believe the scope was modified by mistake in https://github.com/mozilla-services/readinglist/commit/faab111a9d4b204d2fcb5acc8cb1da327cc1d83a. Should we revert it too ?

Natim:
@leplatrem No it is not a mistake it is just the default config that changed.

PR #189, opened 2015-03-18
Adding LICENSE file

### LINKED ISSUE #186: Add LICENSE file
I _assume_ we're releasing this under [Mozilla Public License Version 2.0](https://github.com/mozilla-services/loop-server/blob/master/LICENSE), or something, but it may be nice to have an explicit LICENSE file in the root of the project.

pdehaan:
Fixes #186


Natim:
We should also add a classifier here: https://github.com/mozilla-services/readinglist/blob/master/setup.py#L33

I though we decided to go with APL for server side services.

ametaireau:
Yes, you can find exactly what to do in the cliquet repository, see https://github.com/mozilla-services/cliquet/blob/master/setup.py#L36 and https://github.com/mozilla-services/cliquet/blob/master/LICENSE

Natim:
And https://github.com/mozilla-services/cliquet/blob/master/setup.py#L41

pdehaan:
OK, cool. Let's kill this PR and copy the cliquet license.

pdehaan:
Updated (waiting on Travis)

PR #204, opened 2015-03-20
Add batch operations in load tests

leplatrem:


loadtests/loadtest/__init__.py
```diff
@@ -141,6 +146,20 @@ def read_further(self):
         }
         self._patch(self.random_url, data)
 
+    def batch_read_further(self):
+        data = {}
+        for i in range(25):
+            request = {
+                "path": self.random_url if i % 2 == 0 else self.random_url_2,
```
ametaireau: So it's doing 25 PATCHes on the same two resources, changing the position randomly?

I find this weird :)

Natim: What it is doing is to make a modification 25 on one of the two records selecting them in a random order.

PR #208, opened 2015-03-24
Force a load test action

leplatrem:


leplatrem:
@Natim r?

loadtests/loadtest/__init__.py
```diff
@@ -74,6 +75,11 @@ def test_all(self):
             This method is called as many times as number of hits.
         """
         action, percentage = random.choice(ACTIONS_FREQUENCIES)
+
+        forced_action = os.getenv('LOAD_ACTION')
+        if forced_action:
+            action, percentage = forced_action, 101
```
Natim: 100 would have been enough because randint(0, 100) give you 100 possibilities between 0 and 99

PR #216, opened 2015-03-26
Force default pagination to 100 if not set in settings (fixes #214)

### LINKED ISSUE #214: Better default value for _limit
I currently have 1316 items in my reading list (user: pdehaan), and when I go to https://readinglist.stage.mozaws.net/v1/articles it returns all 1316 items with a `Content-Length` of 674388 (~659 KB)

```
HTTP/1.1 200 OK
Date: Thu, 26 Mar 2015 00:06:31 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 674388
Connection: keep-alive
Last-Modified: 1427328351337
Total-Records: 1316
Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Next-Page, Total-Records, Last-Modified
Backoff: None
Access-Control-Allow-Credentials: true
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Max-Age: 1728000
Access-Control-Allow-Headers: DNT,X-Mx-ReqToken,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Authorization,X-Conditions-Accepted
```

I can set an explicit [`_limit`](https://readinglist.readthedocs.org/en/latest/api/resource.html#paginate) to control how many records I want so I can paginate, but I'm not sure if returning ALL records by default is great.
I know GitHub sets a maximum response of 100 items via their API. Not sure if we want to set a default of 100-200 articles and maybe do some sort of hard limit of 100-200 results per batch.

```js
var _limit = Math.min(parseInt(options._limit, 10), 100);
```
Or whatever the Python syntax would be for that not-very-pretty code.

leplatrem:


readinglist/__init__.py
```diff
@@ -32,5 +32,11 @@ def main(global_config, **settings):
     patch_gevent(settings)
 
     initialize_cliquet(config, version=__version__)
+
+    # Force default pagination (if empty, None or 0)
+    paginate_by = config.registry.settings.get('cliquet.paginate_by')
+    if not paginate_by:
+        config.registry.settings['cliquet.paginate_by'] = 100
```
ametaireau: Which makes me think we should actually have a way for cliquet apps to overwrite the defaults, maybe with a `default.ini` file lying somewhere next the configuration file?

leplatrem: https://github.com/mozilla-services/cliquet/issues/136

PR #219, opened 2015-03-26
Pull monitoring dependencies during install (fixes #218)

### LINKED ISSUE #218: Monitoring dependencies not pulled during install
> <rfkelly> [02:24:43] bobm: looks like you need to install https://pypi.python.org/pypi/statsd
> <bobm> [02:25:36] I thought it might be as simple as that.  Though, I like it when the requirements get pulled in automagically for me.
>

leplatrem:


ametaireau:
Crap, we missed that!

PR #220, opened 2015-03-26
Use existing bearer tokens for smoke testing

ametaireau:


loadtests/config/smoke.ini
```diff
@@ -1,10 +1,6 @@
 [loads]
-users = 10
-duration = 60
```
ametaireau: I took the party to tell that a smoke test isn't ran with multiple users in //, also because we don't have a lot of bearer tokens to use, so 1 user to test it's working should be enough?

leplatrem:
r+

PR #221, opened 2015-03-26
[DO NOT MERGE] Add the werkzeug wsgi middleware profiling

ametaireau:
WIP for profiling.

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -1,6 +1,8 @@
 import pkg_resources
 import logging
 
+from werkzeug.contrib.profiler import ProfilerMiddleware
+
 from pyramid.config import Configurator
 from pyramid.settings import asbool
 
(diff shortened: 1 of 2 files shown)


readinglist/__init__.py
```diff
@@ -39,4 +41,5 @@ def main(global_config, **settings):
         config.registry.settings['cliquet.paginate_by'] = 100
 
     config.scan("readinglist.views")
-    return config.make_wsgi_app()
+    app = config.make_wsgi_app()
```
tarekziade: if that's merged in master - can we have the profiler activated only through a config option ?

ametaireau: yes of  course. The plan is to actually get this in cliquet and not in readinglist, though.

ametaireau:
Fixed in cliquet: https://github.com/mozilla-services/cliquet/commit/50fe4581b8227eb099d86ec5acb1819b2bab9457

PR #227, opened 2015-03-27
Configure sentry logger.

Natim:
This should work according to http://raven.readthedocs.org/en/latest/integrations/pyramid.html#logger-setup


leplatrem:
Please update the docs where we talk about production settings

PR #232, opened 2015-03-30
Use PyFxA for dynamic load testing

almet:
This is a test with PyFxA for dynamic email creations.

Currently, the emails on the restmail service aren't here when it checks and as such it fails. When adding a manual sleep, for whatever raison it claims the assertion is invalid.

@rfk thoughts ?

### DIFF
--- a/loadtests/loadtest/__init__.py
+++ b/loadtests/loadtest/__init__.py
@@ -7,6 +7,10 @@ from requests.auth import HTTPBasicAuth, AuthBase
 from loads.case import TestCase
 from konfig import Config
 
+import fxa.oauth
+from fxa.core import Client
+from fxa.tests.utils import TestEmailAccount
+
 ACTIONS_FREQUENCIES = [
     ('create', 20),
     ('batch_create', 50),
@@ -27,6 +31,8 @@ ACTIONS_FREQUENCIES = [
     ('batch_count', 50),
     ('list_continuated_pagination', 80),
 ]
+
+DEFAULT_FXA_URL = "https://api.accounts.firefox.com"
 
 
 def build_article():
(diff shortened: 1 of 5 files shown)


ckarlof:
@rfk is on PTO this week and OOO at PyCon for the following couple of weeks. @vladikoff?

ametaireau:
@tarekziade I can put that in a `tearDownClass`.

loadtests/loadtest/__init__.py
```diff
@@ -57,18 +63,49 @@ def __init__(self, *args, **kwargs):
...
+            self.user_email,
+            self.random_user
+        )
+
+        # Verify the account using the code from email.
+        self._acct.fetch()
+        for m in self._acct.messages:
+            if "x-verify-code" in m["headers"]:
+                self._fxa_session.verify_email_code(
+                    m["headers"]["x-verify-code"])
```
rfk: The `wait_for_email` helper method may be useful here, it basically encapsulates this loop along with a sleep-and-retry to await email arrival.

loadtests/loadtest/__init__.py
```diff
@@ -57,18 +63,49 @@ def __init__(self, *args, **kwargs):
...
+
+        # Once the account verified, trade the assertion with a bearer token.
+        client = fxa.oauth.Client(client_id=self.conf['fxa-client-id'])
```
rfk: ISTM there may be a potential mis-match between auth-server and oauth-server URLs here, e.g. using the prod auth server with stage oauth server.  Perhaps this is the cause of the "invalid assertion" errors you were seeing?

rfk:
This looks like a good approach, assuming we can get past the "invalid assertion" errors.

ametaireau:
See https://gist.github.com/ametaireau/d3c58857d4e014397281

PR #236, opened 2015-04-01
Use cliquet wsgi middlewares

ametaireau:
Based on top of mozilla-services/cliquet#187

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,6 @@ REQUIREMENTS = [
     'six>=1.9.0',
     'waitress>=0.8.9',
     'cliquet[postgresql,monitoring]>=1.6.0',
-    'raven',
 ]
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -16,7 +16,7 @@
     'six>=1.9.0',
     'waitress>=0.8.9',
     'cliquet[postgresql,monitoring]>=1.6.0',
-    'raven'
```
ametaireau: I'm actually not sure why raven is present here (shouldn't it be directly pulled off from cliquet?)

Natim: Yes it is in the ``cliquet['monitoring']``

Natim: We should remote it.

ametaireau:
This needs cliquet 1.7 (not released yet)

PR #240, opened 2015-04-02
Fix loadtests

ametaireau:


loadtests/loadtest/__init__.py
```diff
@@ -69,6 +71,16 @@ def __init__(self, *args, **kwargs):
...
 
+    def _get_configuration(self):
+        # Loads is removing the extra information contained in the ini files,
+        # so we need to parse it again.
+        config_file = self.config['config']
+        # When copying the configuration files, we lose the config/ prefix so,
+        # try to read from this folder in case the file doesn't exist.
+        if not os.path.isfile(config_file):
+            config_file = os.path.basename(config_file)
+        return Config(config_file).get_map('loads')
```
tarekziade: here I would test again if the file exists before calling Config(), and raise an error if the file is not found. That'll make the problem clearer and != from a file existing but missing a [loads] section

PR #244, opened 2015-04-03
clarified the doc on conflicts

tarekziade:
I also removed the copy-pasted Cliquet behavior description, because the RL is behaving differently

### DIFF
--- a/docs/api/resource.rst
+++ b/docs/api/resource.rst
@@ -265,21 +265,42 @@ an error response is returned with status ``400``.
 Conflicts
 ---------
 
-Articles URL are unique per user (both ``url`` and ``resolved_url``).
+The Reading List Server provides an automatic conflict resolution
+algorithm for articles.
+
+An article is uniquely defined in the database by its ``url`` and
+``resolved_url`` fields. The ``resolved_url`` field can be different
+or the ``url`` value when the client has to follow one or several
+redirections.
 
 :note:
-    A ``url`` always resolves towards the same URL. If ``url`` is not unique, then
-    its ``resolved_url`` won't either.
-
-:note:
-    Unicity on URLs is determined the full URL, including location hash.
+    Unicity on URLs is determined by the full URL, including location hash.
     (e.g. http://news.com/day-1.html#paragraph1, http://spa.com/#/content/3)
 
 :note:
     Deleted records are not taken into account for field unicity.
 
-If the a conflict occurs, an error response is returned with status ``409``.
-A ``existing`` attribute in the response gives the offending record.
+
+When a client pushes an new article which url or resolved url already
+exists in the database,  the automatic conflict resolver will simply keep
+the original one with all its values (title, summary etc.) and return
+to the client its information. No duplicate is created.
+
+Updating the title, excerpt or word_count of an existing article
+won't raise any conflict: the last call wins.
+
+For both updates and creation, you can bypass the automatic
+conflict resolution by adding a **If-Unmodified-Since** in your requests.
+
+In that case, you will get a 412 error if you try to create
+an article with an ``url`` or ``resolved_url`` that already exists
+in the database, or if you try to update an article that has been updated
+by another client in the interim.
+
+Notice that if you add in the same batch two articles with conflicting
+urls, you will get the same behavior as described above, as the batch
+queries are processed in a sequential order.
+
 
 Example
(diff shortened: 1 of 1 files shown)


docs/api/resource.rst
```diff
@@ -265,21 +265,42 @@ an error response is returned with status ``400``.
...
+When a client pushes an new article which url or resolved url already
+exists in the database,  the automatic conflict resolver will simply keep
+the original one with all its values (title, summary etc.) and return
+to the client its information. No duplicate is created.
+
+Updating the title, excerpt or word_count of an existing article
+won't raise any conflict: the last call wins.
+
+For both updates and creation, you can bypass the automatic
+conflict resolution by adding a **If-Unmodified-Since** in your requests.
```
Natim: header in your request.

Please use:

    ``If-Unmodified-Since``

rather than

    **If-Unmodified-Since**

docs/api/resource.rst
```diff
@@ -265,21 +265,42 @@ an error response is returned with status ``400``.
...
+
+In that case, you will get a 412 error if you try to create
```
Natim: That's not exactly true because you will get a 412 also if another article was created for the collection not only if it has the same ``url`` / ``resolved_url`` (In that case you should have 409 as it is the case for cliquet but it is not the behavior chosen for readinglist)

Natim: The only thing you can do is check if the status is 200 and in that case deduce that the article already exists and was not created.

tarekziade: what do you mean by "for the collection" ? the whole database or just the list of articles in your request ?

leplatrem:
I haven't read the whole thing but some clarifications may be relevant in the PATCH docs regarding url modification...

docs/api/resource.rst
```diff
@@ -265,21 +265,43 @@ an error response is returned with status ``400``.
...
+    Unicity on URLs is determined by the full URL, including location hash.
     (e.g. http://news.com/day-1.html#paragraph1, http://spa.com/#/content/3)
 
 :note:
     Deleted records are not taken into account for field unicity.
 
-If the a conflict occurs, an error response is returned with status ``409``.
-A ``existing`` attribute in the response gives the offending record.
+
+When a client pushes an new article which ``url`` or ``resolved_url`` already
```
ametaireau: "a new article"

docs/api/resource.rst
```diff
@@ -265,21 +265,43 @@ an error response is returned with status ``400``.
...
+conflict resolution by adding a ``If-Unmodified-Since`` in your requests
+headers.
+
+In that case, you will get a 412 error if :
+- you are trying to POST a collection and something was changed in the database
+  in the interim.
+- you are trying to POST an article that was changed in the interim.
+
+Notice that if you add in the same batch two articles with conflicting
+urls, you will get the same behavior as described above, as the batch
```
ametaireau: URLs

docs/api/resource.rst
```diff
@@ -265,21 +265,43 @@ an error response is returned with status ``400``.
...
-Articles URL are unique per user (both ``url`` and ``resolved_url``).
+The Reading List Server provides an automatic conflict resolution
+algorithm for articles.
 
-:note:
-    A ``url`` always resolves towards the same URL. If ``url`` is not unique, then
-    its ``resolved_url`` won't either.
+An article is uniquely defined in the database by its ``url`` and
+``resolved_url`` fields. The ``resolved_url`` field can be different
+or the ``url`` value when the client has to follow one or several
```
leplatrem: different from ... (?)

docs/api/resource.rst
```diff
@@ -265,21 +265,43 @@ an error response is returned with status ``400``.
...
+won't raise any conflict: the last call wins.
+
+For both updates and creation, you can bypass the automatic
+conflict resolution by adding a ``If-Unmodified-Since`` in your requests
+headers.
+
+In that case, you will get a 412 error if :
+- you are trying to POST a collection and something was changed in the database
+  in the interim.
+- you are trying to POST an article that was changed in the interim.
```
leplatrem: I'm not it's still the case, but we used the passive form in the docs, without using "you" :)

ametaireau: +1, at least we're trying to :)

Issue #242, opened 2015-04-03
Clarify conflict documentation


tarekziade:
Can we clarify the documentation and explains exactly what happens with these scenarios (what ends in the DB and what gets back to the client body+code) :

= Scenario 1 =

No If-Unmodified-Since header, client A provides a record with an URL that resolves to an URL similar to an URL of a record pushed by client B a few seconds before. But the Client A provide a different title 

Client B:

```json
{
  "defaults": {
    "method" : "POST",
    "path" : "/articles",
    "headers": {
      "Content-Type": "application/json"
    }
  },
  "requests": [
    { "body": {
      "url": "http://example.com/url_1",
      "resolved_url": "http://example.com/real_url",
      "added_by": "Stefan's iPhone",
      "title": "Article 1" } }
  ]
}
```

Then Client A:

```json
{
  "defaults": {
    "method" : "POST",
    "path" : "/articles",
    "headers": {
      "Content-Type": "application/json"
    }
  },
  "requests": [
    { "body": {
      "url": "http://example.com/url_2",
      "resolved_url": "http://example.com/real_url",
      "added_by": "Stefan's iPad",
      "title": "Article Blablba" } }
  ]
}
```

= Scenario 2 =

If-Unmodified-Since header, client A provides a record with an URL that resolves to an URL similar to an URL of a record pushed by client B a few seconds before. But the Client A provide a different title

Client B:

```json
{
  "defaults": {
    "method" : "POST",
    "path" : "/articles",
    "headers": {
      "Content-Type": "application/json",
      "If-Unmodified-Since": "1427464138485",
    }
  },
  "requests": [
    { "body": {
      "url": "http://example.com/url_1",
      "resolved_url": "http://example.com/real_url",
      "added_by": "Stefan's iPhone",
      "title": "Article 1" } },
  ]
}
```

Then Client A:

```json
{
  "defaults": {
    "method" : "POST",
    "path" : "/articles",
    "headers": {
      "Content-Type": "application/json",
      "If-Unmodified-Since": "1427464138485"
    }
  },
  "requests": [
    { "body": {
      "url": "http://example.com/url_2",
      "resolved_url": "http://example.com/real_url",
      "added_by": "Stefan's iPad",
      "title": "Article Blablba" } },
  ]
}
```

= Scenario 3 = 

No If-Unmodified-Since header, client A provides in the same batch two different items that resolves to the same URL but with a different title  

```json
{
  "defaults": {
    "method" : "POST",
[...]

PR #247, opened 2015-04-10
clarified and deduped the data model doc

tarekziade:


leplatrem:
r+

I'll merge #242 first, it looks like it'll conflict since end-line blanks were removed there too

PR #252, opened 2015-04-13
Fixes stored_on not being overwritten at creation (fixes #251)

### LINKED ISSUE #251: stored_on should never be null
https://bugzilla.mozilla.org/show_bug.cgi?id=1153358

leplatrem:


Natim:
Good thanks.

PR #253, opened 2015-04-14
 Rework previous fix about empty titles

leplatrem:
I read again every comment in https://bugzilla.mozilla.org/show_bug.cgi?id=1152915 and had some doubts about the way we fixed the problem.

With the previous fix in #250 
* title field was not required anymore
* empty strings were stored as null

With this fix:
* title is required
* title is stored as null if provided as null
* title is stored as empty string if provided empty

@Natim r?


Natim:
ok it makes sense to me.

PR #256, opened 2015-07-02
Upgrade to Cliquet 2

leplatrem:
I wanted to compare load tests performances now that *Cliquet* has Pypy support. So I started to port *Reading List*.

→ Good exercise, I found [a bug](https://github.com/mozilla-services/cliquet/issues/348) in Cliquet :)

* [x] Upgrade deps
* [x] Update tests payloads and responses
* [x] Clean-up code now available in Cliquet
* [x] Find an elegant fix for https://github.com/mozilla-services/cliquet/issues/348
* [x] Fix tests

### DIFF
--- a/readinglist/__init__.py
+++ b/readinglist/__init__.py
@@ -6,6 +6,7 @@ from pyramid.settings import asbool
 
 import cliquet
 
+
 # Module version, as defined in PEP-0396.
 __version__ = pkg_resources.get_distribution(__package__).version
 
@@ -14,6 +15,11 @@ API_VERSION = 'v%s' % __version__.split('.')[0]
 
 # Main readinglist logger
 logger = logging.getLogger(__name__)
+
+
+DEFAULT_SETTINGS = {
+    'cliquet.paginate_by': 100
+}
 
 
 def patch_gevent(settings):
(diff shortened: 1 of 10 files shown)


readinglist/tests/test_views_article.py
```diff
@@ -150,97 +152,67 @@ def test_read_position_is_reset_when_unread_is_set_to_true(self):
...
+        data = {
             'unread': False,
             'marked_read_by': 'Android',
             'marked_read_on': 543210}
         resp = self.app.patch_json(self.url,
-                                   body,
+                                   {'data': data},
                                    headers=self.headers)
-        self.assertEqual(resp.json['last_modified'],
+        print resp.json['data']
```
Natim: print?

Natim:
We should update the major version readinglist since this break the api (at least the new {"data": {}} formalism)

ametaireau:
That's actually quite interesting to find out what are the breaking changes we did with the new versions of cliquet, and it's quite managable!

Also, this would break the clients because we're now returning the data attribute and using Etags rather than LMS.

leplatrem:
Tests now pass locally, will have to merge and release with this fix though: https://github.com/mozilla-services/cliquet/issues/371

readinglist/__init__.py
```diff
@@ -31,12 +37,8 @@ def main(global_config, **settings):
 
     patch_gevent(settings)
 
-    cliquet.initialize(config, version=__version__)
-
-    # Force default pagination (if empty, None or 0)
-    paginate_by = config.registry.settings.get('cliquet.paginate_by')
-    if not paginate_by:
-        config.registry.settings['cliquet.paginate_by'] = 100
+    cliquet.initialize(config, version=__version__,
```
Natim: :+1:

leplatrem:
> We should update the major readinglist version
Let's do that in a dedicated PR for release

